PR #36, opened 2019-03-20
Fixed a bug in the dataBody when messages value are null.

zthulj:
Had a problem with the view when trying to open a topic with some null values.

Here is a fix for that bug.


tchiotludo:
GG

PR #86, opened 2019-06-21
[WIP] Add a topic regex to basic auth

zthulj:
Hi ! 
I've started to work on this feature. The idea is to provide a way to filter the topics that a user can see when he connects to kafkaHQ.
With this, we will be able to create a dedicated account to each team, with a limited view and rights.
I've been asked for this, and I really thinks it's a usefull one.
Do you agree ?

However, I've got a little problem. I've been able to secure the topic search service. And actually I'm able to detect when opening the topic page that the current user isn't authorized, but as it's a micronaut view, I can't redirect like it's already done (with a HttpResponse.redirect).
Do you have any idea ?

### DIFF
--- a/application.example.yml
+++ b/application.example.yml
@@ -1,8 +1,8 @@
 kafkahq:
   server:
     base-path: "" # if behind a reverse proxy, path to kafkahq with trailing slash (optional). Example: kafkahq is
-                  # behind a reverse proxy with url http://my-server/kafkahq, set base-path: "/kafkahq/".
-                  # Not needed if you're behind a reverse proxy with subdomain http://kafkahq.my-server/
+      # behind a reverse proxy with url http://my-server/kafkahq, set base-path: "/kafkahq/".
+    # Not needed if you're behind a reverse proxy with subdomain http://kafkahq.my-server/
     access-log: # Access log configuration (optional)
       enabled: true # true by default
       name: org.kafkahq.log.access # Logger name
@@ -110,5 +110,4 @@ kafkahq:
         password: pass # Password in sha256
         roles: # Role for current users
           - topic/read
-          - group/read
-          - group/delete
+        topics: ".*" # Regex to filter the topics this user can see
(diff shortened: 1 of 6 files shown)


tchiotludo:
Great idea ! 

To be honest, I think I would have done it by injecting the user directly on class that need to filter topics. 

When you have authentification, you have a method `authentication.getAttributes().get("whatyouwant")`

Example here : https://github.com/tchiotludo/kafkahq/blob/dev/src/main/java/org/kafkahq/controllers/AbstractController.java#L133

The idea is to inject custom attributes on login here : 
https://github.com/tchiotludo/kafkahq/blob/dev/src/main/java/org/kafkahq/modules/BasicAuthAuthenticationProvider.java

`UserDetails userDetails = new UserDetails(auth.getUsername(), auth.getRoles());` will use now this signature :
`public UserDetails(String username, Collection<String> roles, Map<String, Object> attributes) {`

Once you have this, on `TopicRepository` (for example, but needed everywhere you need to filter topics), you can get user / authentification like this : 

```java
    @Inject
    public TopicRepository(KafkaModule kafkaModule, ConsumerGroupRepository consumerGroupRepository, LogDirRepository logDirRepository, ConfigRepository configRepository, ApplicationContext applicationContext) {
        this.kafkaModule = kafkaModule;
        this.consumerGroupRepository = consumerGroupRepository;
        this.logDirRepository = logDirRepository;
        this.configRepository = configRepository;

        if (applicationContext.containsBean(SecurityService.class)) {
            SecurityService securityService = applicationContext.getBean(SecurityService.class);
            Optional<Authentication> authentication = securityService.getAuthentication();
        }
    }
```
Since the authentification can be injected almost everywhere, you can filter all get on topic, and will lead to 404 for unauthorized user globally, no need to go on every controller and view to filter.

What do you think of this way to deal with it ?

zthulj:
Sorry for the delay,
I actually like this way, I'll give it a try soon.

PR #98, opened 2019-07-26
Init helm chart

HungUnicorn:
One issue is `kafkahq-configmap` that is not very flexible to set and it's better to have more in values.yaml, and I would appreciate some suggestions to shift more to `values.yaml`

https://github.com/tchiotludo/kafkahq/issues/96

### DIFF
--- a/helm/chart/.helmignore
+++ b/helm/chart/.helmignore
@@ -1,0 +1,21 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
--- a/helm/chart/Chart.yaml
+++ b/helm/chart/Chart.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for KafkaHQ on Kubernetes
+name: kafkahq
+version: 0.1.0
--- a/helm/chart/README.md
+++ b/helm/chart/README.md
@@ -1,0 +1,25 @@
+# Helm for KafkaHQ
+
+## Install Chart
+To install the KafkaHQ Chart into your Kubernetes cluster :
+
+```bash
+helm install --namespace "kafkahq" --name "kafkahq" kafkahq
+```
+
+After installation succeeds, you can get a status of Chart
+
+```bash
+helm status "kafkahq"
+```
+
+If you want to delete your Chart, use this command:
+
+```bash
+helm delete  --purge "kafkahq"
+```
+
+## Configuration
+Application configuration is in `templates/kafkahq-configmap.yaml`
+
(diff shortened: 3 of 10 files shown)


tchiotludo:
Sorry for delayed response vacations :beach_umbrella: !! 
Thanks for the helm init ! 

To be honest, I haven't done a helm charts for now, so I'm not really aware about the best way to add configuration files.  
Maybe a solution can be to used micronaut feature that enabled every configuration options to be used as env vars like : `KAFKAHQ_CONNECTIONS_{{name}}_PROPERTIES_BOOTSTRAP_SERVERS`.

But I don't like this solution is sexy but it's working. 

Also need a little more time to look at best solution but any help is also welcome !

HungUnicorn:
values.yaml are environment dependent, so stg and pre would use different values and other values should be the same and sit in configmap, like the url of schema-registry and connect.

tchiotludo:
Sorry for the late response.

Maybe this solution solutions  (https://github.com/tchiotludo/kafkahq#docker-1): 
- Use `MICRONAUT_CONFIG_FILES=application.yml,application-common.yml` env var, to have multiple kafkahq configuration files, this allow to have a general one (without brokers configs : application-common.yml) and 1 files with brokers configuration that will be independant by env. 
- set `application.yml` by setting the env var to `KAFKAHQ_CONFIGURATION` with the full yaml with the brokers config like : 
```yaml
kafkahq:
    connections:
        kafka-server: 
            properties:
                bootstrap.servers: ""
                schema-registry:
                    url: ""
                connect:
                    url: "" 
```

You can also have multiple env setting `MICRONAUT_ENVIRONMENTS=prod,helm` env var (same effects than `MICRONAUT_CONFIG_FILES`) for having multiple files that need to be on `/app` path with name `application-helm.yml` for example. 

Hope it will help ?

HungUnicorn:
Thank you for providing feedback. So all the configurations of `application.yml` could be already changed/added in `kafkahq-configmap.yml`. The configmap is transformed into a file called `application.yml`
https://github.com/HungUnicorn/kafkahq/blob/feature_helm_chart/helm/chart/templates/kafkahq-configmap.yaml

Then the file, application.yml, is mounted to the container. 
https://github.com/HungUnicorn/kafkahq/blob/feature_helm_chart/helm/chart/templates/deployment.yaml#L73

The location is notified via `Dmicronaut.config.files=/etc/kafkahq/application.yml`. 
https://github.com/HungUnicorn/kafkahq/blob/feature_helm_chart/helm/chart/templates/deployment.yaml#L41

I add convenient setups for common variables(brokers address), they can be set directly in `values.yaml`, for example
https://github.com/HungUnicorn/kafkahq/blob/feature_helm_chart/helm/chart/templates/kafkahq-configmap.yaml#L63

tchiotludo:
Hello @hchang,

Long time since the start of the pull request, sorry for that.

Some global questions :
Why not put the full yaml of KafkaHQ in values.yaml ?
People will be free to override every config including the charts.

For Kafka server config, I'll prefer to use secrets with full yaml also since there is password in this part.

Let me know what do you think ?
I could also add a commit on your branch to keep your initial work

Like you prefer

HungUnicorn:
Hi @tchiotludo, your idea sounds good and it would be very nice if you add commit(s). I will update the k8s api version to apps/v1 instead of beta.

tchiotludo:
close in favor of #167 

Thanks

PR #110, opened 2019-09-02
Tpc creation conf

ebrard:
- Add configurable retention, replication and partition
- Configurable through the application configuration file (section is maybe not the best one `topic`)
- Add README entry

### DIFF
--- a/src/main/java/org/kafkahq/controllers/TopicController.java
+++ b/src/main/java/org/kafkahq/controllers/TopicController.java
@@ -31,7 +31,6 @@ import org.kafkahq.utils.CompletablePaged;
 import org.reactivestreams.Publisher;
 
 import javax.inject.Inject;
-import javax.validation.Valid;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.time.Instant;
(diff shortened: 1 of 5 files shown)


src/main/java/org/kafkahq/controllers/TopicController.java
```diff
@@ -51,6 +52,12 @@
     private String defaultView;
     @Value("${kafkahq.topic.page-size:25}")
     private Integer pageSize;
+    @Value("${kafkahq.topic.replication:1}")
```
tchiotludo: Remove default value since they are in application.conf (to avoid false duplicate)

src/main/java/org/kafkahq/controllers/TopicController.java
```diff
@@ -31,6 +31,7 @@
 import org.reactivestreams.Publisher;
 
 import javax.inject.Inject;
+import javax.validation.Valid;
```
tchiotludo: Unused import

src/main/java/org/kafkahq/controllers/TopicController.java
```diff
@@ -108,7 +115,10 @@ public HttpResponse list(
     public HttpResponse create(HttpRequest request, String cluster) {
         return this.template(
             request,
-            cluster
+            cluster,
+                "replication", this.replicationFactor,
```
tchiotludo: false ident here

src/main/java/org/kafkahq/controllers/TopicController.java
ebrard: I've removed them from there, but hasn't `page-size` also a default of 25 from the application file ?

ebrard:
@tchiotludo I've addressed the changes, can you please review again ?

tchiotludo:
Awesome, thanks :)

PR #118, opened 2019-09-17
Prevent pollSizePerPartition from returning 0 when size<<partitionsNb

ebrard:
When a topic has significantly more partitions than the defined `topic-data.size`, the function `pollSizePerPartition` may return `0` and kafkhq will instantiate a kafka consumer with a `max.poll.records=0` leading to a 5xx.


tchiotludo:
Good catch :+1: 

Thanks :)

PR #120, opened 2019-09-25
fix connect basic auth

piif:
This change fixes https://github.com/tchiotludo/kafkahq/issues/119


tchiotludo:
Thanks for PR !
To be honest, I don't have a Kafka connect with basic auth, but it seems to be the right way !

piif:
Thanks for the merge ! It works fine on our dev cluster

PR #126, opened 2019-10-09
Added topic regex filter to basic auth users

zthulj:
Hey,

As discussed earlier in the year in #86 , the idea is to provide a way to filter the topics that a user can see when he connects to kafkaHQ.

The implementation is the following : 
If a topics regex is provided in the attributes, the user will be limited to the topics matching this regex. If no regex is provided, the user is able to see everything.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -51,7 +51,7 @@ repositories {
 }
 
 lombok {
-    version = '1.18.8'
+    version = '1.18.10'
     sha256 = ""
 }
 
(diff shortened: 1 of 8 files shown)


tchiotludo:
Looks perfect.👍
Exactly the implementation I was thinking.

Just the mockito part I need to look at to understand why a mock is needed.

And I have added some cache on dev branch, need to look if it work also with cache.

Will come back soon

tchiotludo:
Good work here !! 

Will just add a commit with docs and minor change (name of configuration vars).

Thanks

PR #132, opened 2019-10-15
Fixes #81: Added the version to the kafkaHQ UI from the git property file

### LINKED ISSUE #81: Show version
It would be useful to display the version somewhere in the IHM (perhaps in a corner).
If we don't use docker, there is no information about the deployed version.
Great product !

zthulj:
Hi, 

In order to fix #81,

As we have the tag info into the git.properties, I've used it to display a version info on the ui.
Here is the result (It displays Snapshot if you're not on a tagged branch) :

image

image

Sadly, we can't reuse the GitInfoSource class used by the info endpoint because it has a @Requires(beans = InfoEndpoint.class) annotation. 
The git.properties can't be called just by using a @Value. 

I've watched the GitInfoSource code and made a short version provider that read only one time the tag info and set it.

What do you think ? 
Oh and if you want the tag anywhere else instead of there, just tell me, it's just a proposal

### DIFF
--- a/assets/modules/templates/sidebar.scss
+++ b/assets/modules/templates/sidebar.scss
@@ -15,10 +15,16 @@
         padding: 5px 10px;
         background: $tertiary;
         border-bottom: 4px solid $yellow;
+        position:relative;
         a {
             &:hover {
                 text-decoration: none;
             }
+        }
+        .version{
+            position:absolute;
+            right:5px;
+            bottom:0px;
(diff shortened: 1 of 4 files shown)


tchiotludo:
Seems to be perfect, also try quickly to reuse micronaut endpoint but I can't catch quickly the way ! 
I think It's fine to own GitInfoSource !

PR #139, opened 2019-11-01
Added user visualisation - closes #131

### LINKED ISSUE #131: [feature] Add acls visualization
Hi,

I've started to work a little on this feature cause I wanted it for my cluster. Before to go further, as you made a survey with it, I wanted to know if you're okay with this new feature, and to be sure we agree on the good design on it.

The goal is to add a tab in a topic view, called "Acls".
This tab will be divided in two parts : 

One is to display user acls and the other to display groups acls. In fact, to connect to a topic, a producer need only to have the user correct but a consumer can be limited to one or more groups. 

As I imagine this, I would make it simple, like two tables. One with "user, permission, hosts" and the other with "group, permission, hosts".

I also want to add a new role called "acls/read", because that's a thing we may want to keep for an admin account.

For now, I don't think it's "usefull" to add the modify acls, to start I guess just the visualization is good. 

What do you think ?

zthulj:
In order to implement #131

Added :
- user/read role
- userlist page
- user detail page with topic and group tab
- topic acls tab
- group acls tab


tchiotludo:
Will have a closer look asap.
But in first view, seems to be perfect 👍😋

src/main/java/org/kafkahq/repositories/TopicRepository.java
```diff
@@ -99,11 +103,13 @@ public boolean isListViewMatch(TopicListView view, String value) {
 
     public Topic findByName(String clusterId, String name) throws ExecutionException, InterruptedException {
 
-        Optional<Topic> topics = Optional.empty();
+        Optional<Topic> topic = Optional.empty();
         if(isTopicMatchRegex(getTopicFilterRegex(),name)) {
-            topics = this.findByName(clusterId, Collections.singletonList(name)).stream().findFirst();
+            topic = this.findByName(clusterId, Collections.singletonList(name)).stream().findFirst();
+            topic.ifPresent(t -> t.getUsers().addAll(userRepository.findByResourceType(clusterId, ResourceType.TOPIC, t.getName())));
```
tchiotludo: Seems to be tricky here, you append data to a normally immutable object. 
I prefer to pass data on construct on `Topic` class instead

src/main/resources/views/blocks/acls.ftl
```diff
@@ -0,0 +1,50 @@
...
+                        </h5>
+                    </td>
+                </tr>
+            </#list>
+        </#list>
+
+        <#if aclCounter == 0 >
+            <tr>
+                <td colspan="3">
+                    No ACLS found, or the "authorizer.class.name" parameter is not configured on the cluster.
```
tchiotludo: Surround with `<div class="alert alert-warning mb-0" role="alert"></div>` to be prettier

src/main/resources/views/includes/template.ftl
```diff
@@ -80,6 +80,11 @@
                             <a href="${basePath}/${clusterId}/group"><i class="fa fa-fw fa-object-group" aria-hidden="true"></i> Consumer Groups</a>
                         </li>
                     </#if>
+                    <#if roles?seq_contains("user") == true>
+                        <li class="${(tab == "user")?then("active", "")}">
+                            <a href="${basePath}/${clusterId}/user"><i class="fa fa-fw fa-users" aria-hidden="true"></i> Users</a>
```
tchiotludo: Rename Globally Users with ACLS please to be consistent with Kafka Naming

src/main/resources/views/user.ftl
```diff
@@ -0,0 +1,72 @@
...
+                            </#list>
+                        </h5>
+                    </td>
+                </tr>
+            </#list>
+        </#if>
+        <#if aclCounter == 0 >
+            <tr>
+                <td colspan="3">
+                    No ACL found, or the "authorizer.class.name" parameter is not configured on the cluster.
```
tchiotludo: Same style here :)

src/main/resources/views/userList.ftl
```diff
@@ -0,0 +1,66 @@
...
+                <td>${user.getName()}</td>
+                <td class="khq-row-action khq-row-action-main">
+                    <a href="${basePath}/${clusterId}/user/${user.getEncodedName()}" ><i class="fa fa-search"></i></a>
+                </td>
+            </tr>
+        </#list>
+        <#if users?size == 0 >
+            <tr>
+                <td colspan="2">
+                    No user found, or the "authorizer.class.name" parameter is not configured on the cluster.
```
tchiotludo: Same style here again lol :rofl:

src/test/java/org/kafkahq/repositories/UserRepositoryTest.java
```diff
@@ -0,0 +1,95 @@
...
+    private UserRepository userRepository;
+
+    @Test
+    public void findAll() {
+        assertEquals(2,
+                userRepository.findAll(KafkaTestCluster.CLUSTER_ID, Optional.empty()).size());
+    }
+
+    @Test
+    public void findAll_with_filter() {
```
tchiotludo: just case here and bottom :p 
I hate python, why `_` that remember me python :laughing:

src/main/java/org/kafkahq/repositories/ConsumerGroupRepository.java
```diff
@@ -52,9 +56,11 @@
     }
 
     public ConsumerGroup findByName(String clusterId, String name) throws ExecutionException, InterruptedException {
-        Optional<ConsumerGroup> topics = this.findByName(clusterId, Collections.singletonList(name)).stream().findFirst();
+        Optional<ConsumerGroup> consumerGroup = this.findByName(clusterId, Collections.singletonList(name)).stream().findFirst();
+
+        consumerGroup.ifPresent(cg -> cg.getUsers().addAll(userRepository.findByResourceType(clusterId, ResourceType.GROUP, name)));
```
tchiotludo: Seems to be tricky here, you append data to a normally immutable object.
I prefer to pass data on construct on `ConsumerGroup` class instead

src/main/java/org/kafkahq/modules/KafkaWrapper.java
```diff
@@ -167,4 +170,23 @@ public DescribeClusterResult describeCluster(String clusterId) throws ExecutionE
...
+
+    public Collection<AclBinding> describeAcls(String clusterId, AclBindingFilter filter) throws ExecutionException, InterruptedException {
+        return Lock.call(() -> {
+                    try {
+                        return kafkaModule.getAdminClient(clusterId)
+                                .describeAcls(filter)
+                                .values()
+                                .get();
+                    } catch (ExecutionException e) {
+                        if (e.getCause() instanceof SecurityDisabledException) {
```
tchiotludo:
```
                        if (e.getCause() instanceof SecurityDisabledException | e.getCause() instanceof ClusterAuthorizationException) {
```
If user don't have ACL, it will crash KafkaHQ

src/test/java/org/kafkahq/repositories/UserRepositoryTest.java
zthulj: omg, my brain just randomized tha naming on these ones 😆

src/main/resources/views/includes/template.ftl
zthulj: Just to be dure, can you be more specific on this one ?

- Should I rename the "Users" section in the sidenav to ACLS ? 
- Should I rename roles, class and variables ?
- Should I change the "user" path to "acls" path ?

tchiotludo: Sorry don't see it :-1  and thanks for pointing this :+1: 
 
url / sidebar / role / ClassName & Vars. 

I think there is will be too many people that will mix up with KafkaHQ user.
Since in KafkaHQ, we use the term role, it will be easier to understand I think

zthulj:
The refactor of user has been a bit tricky but I think I came with a nice naming for this feature now.

tchiotludo:
Perfect now !! 
Thanks for the hard work !!!

Issue #143, opened 2019-11-09
[feature] Adding 'groups' to simplify auths


zthulj:
Hey,

While thinking on #142, I thought we could make a change on the way that security configuration is set. It could help to be cleaner I guess even for the basic auth. If you agree, it has to be before starting #142.

What I suggest here is to be able to do this : 

```yml
kafkahq:
  security:
    groups:
      admin: *
      topic_reader: ['topic/read', 'topic/insert', 'topic/delete', 'topic/config/update'
                          'group/read', 'group/delete']

    basic-auth:
      oldWayUser: # Old way user that must still work 
        password: pass
        roles: # Role for current users
          - topic/read
          - topic/insert
          - topic/delete
          - topic/config/update
          - group/read
          - group/delete
        attributes:
          # Regexp to filter topic available for current user
          topics-filter-regexp: "test.*"
       newWayUser: # new user that will use group instead of listing roles
         password: pass 
         group: topic_reader
         attributes:
           topics-filter-regexp: "test.*"
       newWayUserWithAnExtraRole: # new user that will use group instead of listing roles
         password: pass 
         group: topic_reader
         roles:
           - group/read
         attributes:
           topics-filter-regexp: "test.*"
       newAdminUser: # new user that will use group instead of listing roles
         password: pass 
         group: admin
         attributes:
           topics-filter-regexp: "test.*"
```

The goal is to simplify the management of the users in the configuration file. Instead of repeating roles every time, we create the possibility to have group of roles and use them. 

When creating user details, we will merge roles from selected group if there is one, and those from roles list, if they exists. 

What do you think ?

tchiotludo:
Totally agree, but to be totally transparent, peoples working on ldap may have started working on this part, since it more than necessary on ldap auth (group are here in ldap). 

here is the brief I send us : 
```yml
kafkahq:
  security:
    basic-auth:
      user:
        password: d74ff0ee8da3b9806b18c877dbf29bbde50b5bd8e4dad7a3a725000feb82e8f1
        # Change basic auth with roles instead of permission (like now)
        groups:
          - topic-reader
          - admin
​
    # list of groups in kafkaHQ
    groups:
      topic-reader:
        attributes:
           topics-filter-regexp: "test.*"
        roles:       
        - topic/read
      admin:
        roles:
        - topic/read
        - topic/insert
        - topic/delete
        - topic/config/gupdate
        - node/read
        - node/config/update
        - topic/data/read
        - topic/data/insert
[...]

PR #144, opened 2019-11-18
Ldap Suppport with Groups

### LINKED ISSUE #101: LDAP support
I found kafkahq could outperform kafka-manager more, if LDAP support is enabled as kafka-manager does

https://github.com/yahoo/kafka-manager#authenticating-a-user-with-ldap

Is the LDAP support already available via configuration and without any code change? as the micronaut document written here
https://micronaut-projects.github.io/micronaut-security/latest/guide/#ldap

polarising-java:
A first approach to resolve #101.
 Already implemented groups based on #143.


polarising-java:
Add mock tests for Ldap: LdapAuthenticationProviderTest.java

docker-compose.yml
```diff
@@ -105,8 +105,8 @@ services:
 
   kafkacat:
     image: confluentinc/cp-kafkacat
-    command: 
-      - bash 
+    command:
```
tchiotludo: Please remove the whole file modification, not concerning the feature

src/main/java/org/kafkahq/modules/LdapContextAuthenticationMapper.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public AuthenticationResponse map(ConvertibleValues<Object> attributes, String username, Set<String> groups) {
+
+        List<String> kafkaHQgroups = getUserKafkaHQGroups(groups);
+        return new UserDetails(username, userGroupUtils.getUserRoles(kafkaHQgroups), userGroupUtils.getUserAttributes(kafkaHQgroups));
+    }
+
+    /**
+     * Get KafkaHQ Groups configured in Ldap groups
+     * @param ldapGroups
```
tchiotludo: please add comment description or remove `@param` & `@return`

src/main/resources/application.yml
```diff
@@ -30,11 +30,25 @@ micronaut:
     intercept-url-map:
       - pattern: "${kafkahq.server.base-path:}/static/**"
         access: "isAnonymous()"
+     # Ldap configuration
```
tchiotludo: Please remove the whole file modification, not concerning the feature

package-lock.json
```diff
@@ -16,9 +16,9 @@
       "integrity": "sha512-1Bh06cbWJUHMC97acuD6UMG29nMt0Aqz1vF3guLfG+kHHJhy3AyohZFFxYk2f7Q1SQIrNwvncxAE0N/9s70F2w==",
       "dev": true,
       "requires": {
-        "@types/events": "*",
-        "@types/minimatch": "*",
-        "@types/node": "*"
+        "@types/events": "3.0.0",
```
tchiotludo: Please remove the whole file modification, not concerning the feature

tchiotludo:
Can you provide an example configuration that is working with https://www.forumsys.com/tutorials/integration-how-to/ldap/online-ldap-test-server/ for example ? 
I quickly try, and this is not working, must I failed the configuration.

Also, Can you add some configuration example on README.md (maybe with forumsys configurations) ?

src/main/resources/application.yml
polarising-java: Removed changes

package-lock.json
polarising-java: Removed changes

docker-compose.yml
polarising-java: Rollback changes

src/main/java/org/kafkahq/modules/LdapContextAuthenticationMapper.java
polarising-java: Added comments

polarising-java:
I updated the README.md .
Tell me if you want more detail.

tchiotludo:
I've changed `default-roles` with `default-groups` to remove globally roles.
I've made some changes on `README`
Also fixed a bug that won't displayed the login bar if you have only ldap authentification. 

But for now, I just can't make the ldap login work. 
As I quickly understand, the only provider that is used and tested is `BasicAuthAuthenticationProvider` . 
Look like the ldap provider is not used when I tried, maybe can you send me your full configuration in order to find what is wrong on my side ?

polarising-java:
Try with my configuration:
[application.zip](https://github.com/tchiotludo/kafkahq/files/3883526/application.zip)

users (username/password - group):
riemann/password - mathematicians
einstein/password - scientists

First goes to BasicAuthAuthenticationProvider, and if no mtach then goes to LdapAuthenticationProdvider.

tchiotludo:
ok !!! found it !! 
I'm a super Nooob !! :no_entry: 
Yaml space failed :) missing 2 one on ldap part ... 

Everything seems ok for me. 
If you just can confirm that my last modification (with groups) seems ok for you and you can merge it :fast_forward: :+1: :1st_place_medal:

polarising-java:
Found also 2 situations:

- The default-groups doesn't apply the topics-filter-regexp

- if i remove default-groups from the configuration crashes with "Message: Error resolving field value [${kafkahq.security.default-groups}]. Property doesn't exist or cannot be converted"

tchiotludo:
* default-groups / roles never allow `topics-filter-regexp` has I know ? I'm wrong ?  
* remove default-groups should not failed since I've added in the main configuration here 

https://github.com/polarising-public/kafkahq/blob/ce1a9d216e7bc06796e1b159adbbd102951e253b/src/main/resources/application.yml#L108

Can you send me the snippets that crash the app ?

polarising-java:
If you remove default-groups from security section like this:

`security:
    groups:
      admin:
        roles:
        - topic/read
        - topic/insert
        - topic/delete
        - topic/config/update
        - node/read
        - node/config/update
        - topic/data/read
        - topic/data/insert
        - topic/data/delete
        - group/read
        - group/delete
        - group/offsets/update
        - registry/read
        - registry/insert
        - registry/update
        - registry/delete
        - registry/version/delete
        - acls/read
        - connect/read
        - connect/insert
        - connect/update
        - connect/delete
        - connect/state/update
      reader:
        roles:
          - topic/read
          - node/read
          - topic/data/read
          - group/read
          - registry/read
          - acls/read
          - connect/read`

tchiotludo:
really strange, I can't reproduce on my side.
The default `application.yml` is used and It works for me. 
Do you have update with my last modification on this file ?

polarising-java:
"default-groups / roles never allow topics-filter-regexp has I know ? I'm wrong ?"

You are right, but since we have default groups instead of default roles, and in a Group you can define a topics-filter-regexp, it's quite strange not applying the topic filter in the case of a default group.

parmarques:
Added the topic filter regex also to default groups

PR #155, opened 2019-11-26
Build the search URL using segments

ankon:
This fixes problems with the missing encoding of the path parameter, and at the same time avoids doing
string manipulation on things we know already what they are ("URL path segments"). URI.js does a good
job of combining these as needed already.

---
Reproduction case: Try search for something like `user/created`, which should produce a URL looking like `.../search/user%2fcreated`. Before this fix it would produce `.../search/user/created`, which (rightfully) returns a 401/404.


tchiotludo:
good catch and good fix ! :+1: 

thanks

PR #163, opened 2019-12-04
Add La Redoute

DanielOlival:
In La Redoute we are also using kafkahq, so we want to add La Redoute to the Who's using KafkaHQ section


tchiotludo:
thanks :+1:

PR #165, opened 2019-12-05
change fonts regex to be windows compatible

### LINKED ISSUE #153: building on windows
Is there a way to build kafkahq on windows, pretty much all of the assemble targets i run error out with

```
> Task :webpack FAILED
C:\src\kafkahq\node_modules\.bin\webpack:2
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
          ^^^^^^^

SyntaxError: missing ) after argument list
    at Module._compile (internal/modules/cjs/loader.js:895:18)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:995:10)
    at Module.load (internal/modules/cjs/loader.js:815:32)
    at Function.Module._load (internal/modules/cjs/loader.js:727:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1047:10)
    at internal/main/run_main_module.js:17:11
```

I know nothing about webpack, but it looks to be a platform tooling issue? I verified and both `dirname` and `sed` are available in the path using gnutools.

jorgheymans:
fixes #153

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -132,7 +132,7 @@ module.exports = (env, argv) => {
                 {
                     test: /\.(jpe?g|png|gif|svg)$/,
                     exclude: [
-                        /\/fonts\/|\\fonts\\/
+                        /[\\/]fonts[\\/]/
                     ],
                     loader: 'file-loader',
                     options: {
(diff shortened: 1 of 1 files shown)


tchiotludo:
Oh my god ! 
Is it a real fix that you have tested ? 
Really think that webpack handle this transparently ...

jorgheymans:
well it runs the build to completion, successfully. Whether it produces the correct application archive is another thing maybe, did not test i admit.

jorgheymans:
Also maybe this https://github.com/FortAwesome/react-fontawesome would avoid such issues instead of using fontawesome directly ?

tchiotludo:
Really surprised but It's true !! 
look at https://webpack.js.org/plugins/split-chunks-plugin/#optimizationsplitchunks

Change changed to recommended webpack notation 

Thanks

PR #167, opened 2019-12-11
Inititalize helm template

snahelou:
Hello

This is the first version of a helm chart to deploy your awesome app in Kubernetes using configmap and secret.

Feel free if you wand improvements :) 

BR

Tested with helm3


tchiotludo:
Almost perfect, will only change some variable name to be perfect 

Thanks a lot !!

Issue #127, opened 2019-10-10
Connect Multicluster


alpeza:
Hello

We are going to make that the KafkaHQ can interact with multiple Connect clusters. We would like make a pull request later. Did you think of any concrete way to do this and we would stick to this pattern?

Regards

tchiotludo:
Just to be sure to understand ?
You want to allow multiple connect cluster for 1 Kafka cluster ?

alpeza:
exact.
and thanks for the reply

alpeza:
Hi this is what we need:

![](https://i.pinimg.com/originals/78/5d/de/785dde837f45bdb3d0f693cef43c7866.png)


We have thought two approaches for implement this:

### First Approach


We put in the `application.yml` file the urls of the different connect clusters.

![](https://i.pinimg.com/originals/cd/dc/41/cddc4116186676414fa02a09c1065684.png)

### Second Approach

Another approach is that KafkaHQ discovers the Connects that are added to the Kafka Broker.
This can be do by capturing the topics that Connect generates internally.


![](https://i.pinimg.com/originals/70/23/25/702325d18543b76be80b1958a1edf554.png)


We would like to add one or another of these solutions to the main version of HQ. Would you let us implement any of these solutions to add to the master version

thanks

tchiotludo:
Hello, 

Thanks for your detail review 👍

I'll vote for first version. I think it will work that all the use case (for example, dedicated cluster for storing offsets). 
The second one is a little big more tricky, and maybe not work if there is an authentication on connect cluster (that is the case for most saas Kafka provider)

For the list of connect cluster list, I prefer to have the list of cluster on the left menu (like the Kafka cluster), and to have different route for the different connect endpoint, so configuration of connect must have a name (not a simple array in your example)

Let me know if it's ok for you or you need more precision

And thanks to improve KafkaHQ 👍

alpeza:
Hi tchiotludo

Functionally I understand something like this:


* `Application.yml`:

```yaml
  connect-clusters:
    MyConnectCluster1:
      url: "http://thednsofmyconnectcluster:8083"
      basic-auth-username: "connect basic auth username"
      basic-auth-password: "connect basic auth password"
      ssl-trust-store: "/app/truststore.jks"
      ssl-trust-store-password: "trust-store-password"
      ssl-key-store: "/app/truststore.jks"
      ssl-key-store-password: "key-store-password"
    YetAnotherConnect:
[...]

PR #170, opened 2019-12-17
Connect Multicluster #127

polarising-java:
#127 Connect Multicluster development.

Transform single connect to a list of connects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,8 @@ file example can be found here :[application.example.yml](application.example.ym
     * `url`: the schema registry url 
     * `basic-auth-username`: schema registry basic auth username
     * `basic-auth-password`: schema registry basic auth password
-  * `connect`: *(optional)*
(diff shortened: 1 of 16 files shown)


tchiotludo:
Great work !!! 
Works well, all is perfect ! 

Thanks a lot

ankon:
How does this affect existing configurations?

tchiotludo:
It only changed the configuration of connect to an array instead of object : 
https://github.com/tchiotludo/kafkahq/blob/c09be06d175104f4a3509d2a0566f7be781406d8/application.example.yml#L45-L65

ankon:
Right. If I understand correctly then this would break existing configurations, which would need to be changed to an array of named things as well. Would be good to mention that in the release notes :)

tchiotludo:
yes right, forgot to add it, just corrected thanks :)

ankon:
Thanks. 

Btw, @tchiotludo did you see my other invitation/issue? :)

Issue #150, opened 2019-11-21
Some utf8 characters not rendered correctly on windows


jorgheymans:
In the topic overview, it seems that the ≈ character is not rendered correctly. We're using chrome or FF on windows 7.

![image](https://user-images.githubusercontent.com/193792/69340113-04357b00-0c67-11ea-8156-4663b4cf328e.png)

topicList.ftl:
```html
                            <span class="text-nowrap">
                                ≈ ${topic.getSize()}
                            </span>

```

jorgheymans:
Also similar issues in the login screen:

![image](https://user-images.githubusercontent.com/193792/69340857-b1f55980-0c68-11ea-8029-2f3e111ea0ca.png)

Seems to come from the CSS: 

![image](https://user-images.githubusercontent.com/193792/69340889-c2a5cf80-0c68-11ea-8544-56ba14377566.png)

jorgheymans:
All using latest release, started via `java -jar ...`

jorgheymans:
sorry for spamming this issue but it's weird, after a restart the issues with the login screen are gone. Only the size column character issue ≈ remains. Not sure what is going on :/

tchiotludo:
Can you try with a simple curl on topic list page ? 
The ≈ should be ok normally ? 
(to be sure it's not a browser issue)

jorgheymans:
Also using curl the char is not rendered correctly:

```
                        <td>
                            <span class="text-nowrap">
                                â‰ˆ 855,221
                            </span>
                        </td>
                        <td>
                                193.4 mo
                        </td>
                        <td>20</td>
                        <td>5</td>
                        <td><span class="">5</span></td>
                        <td>
```

jorgheymans:
You can mark this as a windows issue, just verified that the issue disappears as soon as we deploy on a linux server.

tchiotludo:
ok, the bug on windows ... 
Hope some one with windows could help :)

goncaloValentim:
I think I found the cause of this issue!

The ftl templates are not setting the encoding as UTF-8, which results in using the system's default encoding (iso-8859-1 in Windows 10) . This causes UTF-8's chars not rendering properly. By setting the "encoding" parameter in each template I was able to fix this issue:

`<#ftl output_format="HTML" encoding="UTF-8">`

![image](https://user-images.githubusercontent.com/28408419/71004393-cd257e80-20d9-11ea-90b9-3405559aa8cf.png)


[...]

PR #172, opened 2019-12-17
Added encoding notation to templates in order to support UTF-8

goncaloValentim:


tchiotludo:
Thanks :) I really think it was default behaviour 

Must be close also #150

PR #186, opened 2020-01-15
Add a reference to NEXT Technologies

ankon:
Using KafkaHQ for development/support purposes in a microservice environment.

Kafka setup is AWS MSK, with a few hundred topics, and countless consumer groups.


tchiotludo:
thanks :+1:

PR #191, opened 2020-01-17
Indent extraVolumeMounts to same level as secrets/configuration mounts

### LINKED ISSUE #190: Incorrect indentation of extraVolumeMounts in Helm chart
When trying to use the Helm chart to deploy an instance of KafkaHQ that has some extra volumes mounted (in my case, a truststore), the indentation of the extra volumes section of the generated deployment is wrong.

For example, using a values file that has both `secrets` and `extraVolumeMounts` defined, something like:

```
secrets: |
  foo

extraVolumeMounts:
  - name:      foo
    mountPath: /foo
    readOnly:  true
```

then the Helm chart will produce a `deployment.yml` that looks like this:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  [...]
spec:
  [...]
  template:
    [...]
    spec:
      containers:
        - name: kafkahq
          [...]
          volumeMounts:
          - mountPath: /foo
            name: foo
            readOnly: true
            - name: secrets
              mountPath: /app/application-secrets.yml
              subPath: application-secrets.yml
          [...]
```

i.e. the indentation of the custom extra `volumeMounts` is less that of the `secrets` volume.

This is probably because _either_ 10 or 12 spaces of indentation would work here (because YAML allows the `-` of an array to be unindented or indented) but it must be consistent within a block.

I'll make a PR to suggest changing the indentation in the following line:
https://github.com/tchiotludo/kafkahq/blob/402c3e94e053eb811055c31d13026cb7458d1485/deploy/helm/templates/deployment.yaml#L51

dominics:
Fixes #190

### DIFF
--- a/deploy/helm/templates/deployment.yaml
+++ b/deploy/helm/templates/deployment.yaml
@@ -48,7 +48,7 @@ spec:
 {{ toYaml . | trim | indent 12 }}
           {{- end }}
           volumeMounts:
-          {{- if .Values.extraVolumeMounts }}{{ toYaml .Values.extraVolumeMounts | trim | nindent 10 }}{{ end }}
+          {{- if .Values.extraVolumeMounts }}{{ toYaml .Values.extraVolumeMounts | trim | nindent 12 }}{{ end }}
           {{- if .Values.configuration }}
             - name: config
               mountPath: /app/application.yml


tchiotludo:
Nice !
Thanks

PR #199, opened 2020-01-24
Show a question circle icon if the type of the connector is not present

saparicio:
If the connect API does not return the type of the connector (this can happen with old kafka connect deployments" an error was shown in the browser


tchiotludo:
I can really try, so I trust you ! 
Thanks :+1:

PR #205, opened 2020-01-31
Use more common units and names for expressing file size

ankon:
* Use 'Size' instead of 'Weight' in the topic list UI, and rename the existing column to 'Count'
* Use SI binary units (B, Kibibyte, Mebibyte, ...) instead of the 1000/octet-based numbers

---
Expressing a file "weight" in "octets" is quite common in the French locale, but this is rather confusing for English -- where it is a file "size" in "bytes". This commit tries to change to the binary SI units as per https://en.wikipedia.org/wiki/Byte.


tchiotludo:
agree with that :) 
Maybe my french origin allow me to talk [Franglais](https://en.wikipedia.org/wiki/Franglais)

PR #219, opened 2020-02-12
Role assignments to individual LDAP users

KarstenWintermann:
Role assignments to individual LDAP users makes it possible to use LDAP purely for authentication, when defining LDAP groups is not wanted.

Role assignments from groups are merged together with individual role assignments. 

Description is added in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,6 +411,7 @@ Dev server is a java server & webpack-dev-server with live reload.
 * [Adeo](https://www.adeo.com/)
 * [Auchan Retail](https://www.auchan-retail.com/)
 * [Bell](https://www.bell.ca)
+* [BMW Group](https://www.bmwgroup.com)
 * [Boulanger](https://www.boulanger.com/)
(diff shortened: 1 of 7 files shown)


tchiotludo:
Hi, 
Good idea, it will help all the people that just can't map group because internal AD is messy ! 
I also think it will help people to quick start with ldap (when i see the list of trouble they have with it)

Some modification are welcome:  
* Add a example [here](https://github.com/tchiotludo/kafkahq/blob/dev/application.example.yml), i'll try to keep this file will all possibility 
* Maybe add a unit test can be safe [here](https://github.com/tchiotludo/kafkahq/blob/dev/src/test/java/org/kafkahq/modules/LdapAuthenticationProviderTest.java) to cover the modification

KarstenWintermann:
Thanks for your comments. I added the example and the unit test.
Also, I added BMW Group to the list of users. Very useful tool!

tchiotludo:
Seems to be perfect  :) 

Thanks for the good work !

PR #228, opened 2020-02-21
WIP: Produce with schema

francoisblarel:
TODO :
- [x] add unit test
- [x] refactoring : should we create a dedicated Record object for the production? or merge with the actual Record?

### DIFF
--- a/src/main/java/org/kafkahq/models/Record.java
+++ b/src/main/java/org/kafkahq/models/Record.java
@@ -13,6 +13,7 @@ import java.nio.ByteBuffer;
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 
 @ToString
 @EqualsAndHashCode
--- a/src/main/java/org/kafkahq/models/Record.java
+++ b/src/main/java/org/kafkahq/models/Record.java
@@ -13,7 +13,6 @@ import java.nio.ByteBuffer;
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Optional;
 
 @ToString
 @EqualsAndHashCode
--- a/src/main/resources/views/topicProduce.ftl
+++ b/src/main/resources/views/topicProduce.ftl
@@ -1,6 +1,8 @@
 <#ftl output_format="HTML" encoding="UTF-8">
 
 <#-- @ftlvariable name="topic" type="org.kafkahq.models.Topic" -->
+<#-- @ftlvariable name="schemasList" type="java.util.List<org.kafkahq.models.Schema>" -->
+
 
 <#import "includes/template.ftl" as template>
 <#import "includes/functions.ftl" as functions>
@@ -22,9 +24,21 @@
     <div class="form-group row">
         <label for="key-schema" class="col-sm-2 col-form-label">Key schema id</label>
         <div class="col-sm-10">
-            <input type="text" class="form-control" name="keySchema" id="key-schema" autocomplete="off" placeholder="Key Schema id">
+            <select
+                    name="keySchema"
+                    id="key-schema"
+                    class="khq-select form-control col-sm-4"
+                    data-style="btn-white"
+                    data-live-search="true"
+                    title="Schemas"
+            >
+                <#list schemasList as schema>
+                    <option value="${schema.getId()}">${schema.getSubject()}</option>
+                </#list>
+            </select>
         </div>
     </div>
+
     <div class="form-group row">
         <label for="key" class="col-sm-2 col-form-label">Key</label>
         <div class="col-sm-10">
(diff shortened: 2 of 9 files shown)


src/main/java/org/kafkahq/models/Record.java
```diff
@@ -13,6 +13,7 @@
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
```
tchiotludo: remove this noise please :)

src/main/resources/views/topicProduce.ftl
```diff
@@ -19,6 +19,12 @@
             </select>
         </div>
     </div>
+    <div class="form-group row">
+        <label for="key-schema" class="col-sm-2 col-form-label">Key schema id</label>
+        <div class="col-sm-10">
+            <input type="text" class="form-control" name="keySchema" id="key-schema" autocomplete="off" placeholder="Key Schema id">
```
tchiotludo: Please change this to a combo box like here : 
https://github.com/tchiotludo/kafkahq/blob/fdac9438b7c64bab1e1905dbe7dcc69e23caf4f3/src/main/resources/views/tail.ftl#L36-L45
with the list of available schema on current registry instead of id

src/main/resources/views/topicProduce.ftl
```diff
@@ -35,12 +41,19 @@
...
         </div>
     </div>
+
     <div class="form-group row">
         <label for="timestamp" class="col-sm-2 col-form-label">Timestamp</label>
         <div class="col-sm-10 khq-datetime">
             <input type="text" class="form-control datetimepicker-input" name="timestamp" id="timestamp" autocomplete="off" data-toggle="datetimepicker" data-target="#timestamp" placeholder="Timestamp"/>
         </div>
     </div>
+    <div class="form-group row">
```
tchiotludo: same than below :smile:

tchiotludo:
Good stuff here :) 
Thanks :+1:  

Just the combo box with available schema will be very nice.
Also don't understand the ` dedicated Record object for the production`, can you tell me more ?

src/main/resources/views/topicProduce.ftl
francoisblarel: No problem. 
Do you want all the schemas or just the last version of each subject?

tchiotludo: I would say the last version of the schema, dont think people use outdated schema most of the time. 
Also if it's possible to sort the list and display the corresponding schema on top from this topic (example topic name: my-topic / display on top my-topic-value or key schema) ?

francoisblarel:
> Also don't understand the ` dedicated Record object for the production`, can you tell me more ?

At first, I wanted to re-use the Record entity, and fill all the attributes to finally call a "serialize" method on it, but the use case is different of the actual Record which is more or less a wrapped ConsumerRecord. So I wondered if I should have created some kind of ProducerRecord wrapper.
As a first step I only created a dedicated serialization class (maybe its sufficient)

src/main/resources/views/topicProduce.ftl
francoisblarel: Alright. Great idea for the sorting :+1: (although only useful when using topic-named strategy)

tchiotludo:
ok better understanding now :) 
dedicated serialization class don't shock me, you can keep it as I think ! 
the modules directory is little messy but one day, i will cleanup it and separate some of them to a `services` directory, but it's fine for now :)

src/main/resources/views/topicProduce.ftl
tchiotludo: I think most of people use topic-named strategy and will be happy.
Others just have to search in the combo :laughing:

tchiotludo:
Seems to be better now ! 
Thanks

Last things I see is ACLS, if the user has no right on the schema registry and there is no configured schema registry, you should not displayed this feature.
I think for now, no SR mean a null pointer exception

francoisblarel:
You're right, I will check this next week when coming back from hollidays :D

tchiotludo:
thanks :) 

but for ACLS, I mostly think about this conditions : 
https://github.com/tchiotludo/kafkahq/blob/2e2cde3b85114529484b4af92ccade7a93d39dfa/src/main/resources/views/includes/template.ftl#L90
That is available globally for all templates from : 
https://github.com/tchiotludo/kafkahq/blob/2e2cde3b85114529484b4af92ccade7a93d39dfa/src/main/java/org/kafkahq/controllers/AbstractController.java#L74

It will check if registry acls and configuration are present ;)

francoisblarel:
Sorry I missed this point -_-'
I'will have a check on this :)

tchiotludo:
All good work :) 
It's merged and I just add a little commit on layout.

Thanks a lot :)

PR #233, opened 2020-02-27
Add the possibility to have a varenv of jvm.options path

clallavena:
Modify kafkahq script to allow it to accept a varenv of JVM_OPTS_FILE, make the configuration templating easier with docker-compose for example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@
     - [Stand Alone](#stand-alone)
     - [Kubernetes using Helm](#running-in-kubernetes-using-a-helm-chart)
 - [Configuration](#configuration)
+    - [JVM.options file](#run-with-another-jvm-options-file)
     - [Kafka cluster](#kafka-cluster-configuration)
     - [KafkaHQ](#kafkahq-configuration)
     - [Security](#security)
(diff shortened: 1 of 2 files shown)


docker/app/kafkahq
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env sh
 
 # Read user-defined JVM options from jvm.options file
-JVM_OPTS_FILE=${JVM_OPTS_FILE:-/app/jvm.options}
+if [ -z "${JVM_OPTS_FILE}" ]; then
```
tchiotludo:
```
# JVM_OPTS_FILE=${JVM_OPTS_FILE:-/app/jvm.options}
# echo $JVM_OPTS_FILE
/app/jvm.options
# JVM_OPTS_FILE=${JVM_OPTS_FILE:-/app/jvm.options}^C
# export JVM_OPTS_FILE="/tmp/bla"
# JVM_OPTS_FILE=${JVM_OPTS_FILE:-/app/jvm.options}
# echo $JVM_OPTS_FILE
/tmp/bla
```

This `if` is exactly the same than the previous one `${JVM_OPTS_FILE:-/app/jvm.options}` that will take the default parameters only if `JVM_OPTS_FILE` is not defined

README.md
```diff
@@ -141,6 +142,39 @@ helm install --name=kafkahq-release-name  .
 Configuration file can by default be provided in either Java properties, YAML, JSON or Groovy files. YML Configuration
 file example can be found here :[application.example.yml](application.example.yml)
 
+### Run with another jvm.options file
+
+By default, the docker container will run with a [jvm.options](docker/app/jvm.options) file, you can override it with
+your own with an Environment Variable. With the `JVM_OPTS_FILE` varenv, you can override the jvm.options file by passing
```
tchiotludo:
```suggestion
your own with an Environment Variable. With the `JVM_OPTS_FILE` environment variable, you can override the jvm.options file by passing
```

PR #242, opened 2020-03-06
Add Klarna as KafkaHQ user

RickardCardell:
We at Klarna are using KafkaHQ so I'm adding us as an official user of this awesome GUI!


tchiotludo:
thanks :+1:

PR #269, opened 2020-04-20
Remove the unneeded `node_modules/.bin/` path in npm scripts

ankon:
npm will put that directory explicitly on the PATH, and so there is no point in having these here.


tchiotludo:
good catch, thanks @ankon !

PR #270, opened 2020-04-20
Use the JRE variant of the openjdk image

ankon:
There is no reason really to use a full JDK image for running akhq, and switching to the JRE variant saves ~150MiB.

---
FWIW: 7114c33 changed from a JRE image to a JDK image, but it isn't obvious why -- and the image built using the JRE image seems to work just fine.


tchiotludo:
lool :) 
what a noob I am :) 

Thanks ;)

PR #273, opened 2020-04-21
Redirect to /login/unauthorized on authorization errors

### LINKED ISSUE #268: akhq doesn't redirect to /login for unauthenticated user
Initially just a comment, but it seems a new issue is better here as the problem is slightly different: https://github.com/tchiotludo/akhq/issues/261#issuecomment-614667294

There is still a problem that opening opening the root of the akhq instance, say "https://akhq.internal.example.com" shows the exception (after redirecting to the /cluster/topic endpoint, see #267).

On the "dev" version I can login when opening "https://akhq.internal.example.com/login", and it seems that afterwards akhq works as expected, on 0.14.0 that doesn't work.

ankon:
This fixes #268, which basically was caused by the error controller "eating" the authorization error, so that the actual session configuration for how to redirect for authorization errors didn't get used anymore.

The proposed fix limits the error controller to only show the internal error page for actual internal errors -- but my micronaut knowledge isn't good enough to know whether this is the correct way.

### DIFF
--- a/src/main/java/org/akhq/controllers/ErrorController.java
+++ b/src/main/java/org/akhq/controllers/ErrorController.java
@@ -49,7 +49,7 @@ public class ErrorController extends AbstractController {
             .body(error);
     }
 
-    @Error(global = true)
+    @Error(status = HttpStatus.INTERNAL_SERVER_ERROR, global = true)
     public HttpResponse<?> error(HttpRequest<?> request, Throwable e) {
         log.error(e.getMessage(), e);
 


tchiotludo:
This one will need some time to be sure about that.
Will look at this week

tchiotludo:
hey @ankon, I think last commit is the proper way to handle this.
Can you try please ?

jorgheymans:
@tchiotludo works for us, thanks !

ankon:
I tried 0.14.1 now, and it works just fine for us, thanks!

I'll close this one then.

PR #274, opened 2020-04-21
Use temporary redirects for / -> /{cluster}/topic

ankon:
The configuration name for the cluster can change easily, so we need a temporary redirect to prevent the browser from caching the redirection in its own cache, and then always going to the wrong place.


tchiotludo:
also a good idea, that is annoying ! 
don't think I could do a permanent one here , ... 
:+1:

PR #275, opened 2020-04-21
Do not send the 'node' directory to docker

ankon:
This directory contains the NodeJS runtime for Gradle _when building the frontend assets_, but it isn't
needed when packaging the docker image.

---
This should have no impact at all on the runtime, but speed up builds a bit.


tchiotludo:
good idea :)

Issue #285, opened 2020-05-02
Viewing Connector Details throws 500


tjharke13:
When viewing a connector details a 500 error is thrown.  Looking through source code and looking at kafka connect logs, the issue stems from the getPlugin calling getPlugins and trying to validate all plugins.  Even plugins that are deployed, but not in use.  This causes a 500 error on kafka connect side for a NPE when trying to validate a plugin that isn't deployed for a connector.

Because this should be doing a singular getPlugin, there are unneeded calls being done out to kafka connect to validate the plugins.  

![image](https://user-images.githubusercontent.com/3589858/80864790-81b7db00-8c4a-11ea-8c8d-3ec31038d411.png)

tchiotludo:
closed in #287

PR #287, opened 2020-05-02
Change functionality of getPlugin to not get all plugins and try vali…

tjharke13:
…dating them.  This causes a NPE for connectors that aren't in use and deployed as connectors.  This will also improve performance as it will no longer validate all connector plugins, but only the one necessary.


tjharke13:
#285

tchiotludo:
Seems to be good, thanks :+1:

PR #289, opened 2020-05-12
Fixed issue where keystore was not being set properly for Kafka Connect

CodyFearer:
Fixed issue where keystore for Kafka Connect was being set to the truststore.


tchiotludo:
Thanks :+1:

PR #301, opened 2020-05-30
Adjusted replica count

xpr3sso:
Modified the approach for getting replica and in-sync replica count

### DIFF
--- a/src/main/java/org/akhq/models/Topic.java
+++ b/src/main/java/org/akhq/models/Topic.java
@@ -73,19 +73,16 @@ public class Topic {
 
     public long getReplicaCount() {
         return this.getPartitions().stream()
-            .flatMap(partition -> partition.getNodes().stream())
-            .map(Node::getId)
-            .distinct()
-            .count();
+            .mapToLong(partitions -> partitions.getNodes().size())
+            .max()
+            .orElseGet(() -> 0L);
     }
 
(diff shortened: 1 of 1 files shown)


tchiotludo:
Can you help me a little bit ?
Is this just style or fixing a wrong value ?

xpr3sso:
Yes, ofc. I encountered this when testing a topic (10 partition, replication 1) on a 4 broker cluster, returning replication factor and in-sync relpica of 3. The code was checking for unique nodes the topic's partitions were distributed on, in my case broker-0 to broker-2 (borker-3 on standby).
So e.g. partiton-0 was on broker-0 only, partition-1 on broker-1 only and partition-2 on broker-2 only. This resulted in the mentioned value of 3, which was really only 1 in my case.

tldr ;)
The approach in this PR is, to give an optimistic value for the replica count (max. value of nodes for a partition) and a pessimistic one for in-sync replicas (min. value of nodes found for a partition). So when the cluster is healthy both values should express the configured topic values, if one of the brokes is down on the other hand, in-sync replicas will adjust accordingly.

tchiotludo:
That make sense !
Thanks for clarifying and for the fix 👍

xpr3sso:
Glad to help.
Thanks for your great tool :)

tchiotludo:
Thanks a lot 👍

PR #315, opened 2020-06-19
add initcontainers value to the helm chart

skhedim:
Add value to attach an initcontainer to the akhq deployment. I need this to generate the keystore.

PR #317, opened 2020-06-23
add initcontainers value to the helm chart

skhedim:
Same as PR #315, I just change the destination branch to master to trigger the  helm chart build in github actions.


tchiotludo:
maybe just increment the helm charts version, no ?

tchiotludo:
Thanks ! 
Maybe a should add a note to tell that PR for helm must be done on master :)

PR #324, opened 2020-07-02
Add topic configuration button in topic list screen

RajKumarMondol:
**Problem:**
Needed configure button/action to go to the topic's configuration. 
This is very much useful when we want to view or modify topic's configuration. Without this feature, we have to wait for the topic to read data(default action) before we can go to the config tab.

**Action**
Add topic's configuration button in topic list ui.


tchiotludo:
Make senses :) 

Thanks

PR #334, opened 2020-07-10
Display consumer groups ordered by group id

RajKumarMondol:
**Problem:**
Consumer Groups page displays consumers in random order. 
When this list grows it is much helpful to navigate if these consumer groups are sorted by Id.

**Action**
Return consumer group list after sorting it by consumer group id from ConsumerGroupRepository


tchiotludo:
Never have the issue (really think it's sorted by Kafka) but can't be worst ;)

PR #337, opened 2020-07-19
Sort topic records by timestamp across all partitions

RajKumarMondol:
**Problem**
While sorting topic records by "**Newest**" or "**Oldest**". Records are getting sorted by timestamp only withing individual partitions. 

- Sorted by "Newest" 
image

- Sorted by "Oldest"
image

- When the number of partitions is more. The user feels a few records are not visible or missed by akhq. 
- The user spends more time to find the desired records and increases his analysis time

**Actions**
Records to be sorted by record timestamp across all partitions.

- Sorted by "Newest" After implementation
image

- Sorted by "Oldest" After implementation
image

### DIFF
--- a/src/main/java/org/akhq/repositories/RecordRepository.java
+++ b/src/main/java/org/akhq/repositories/RecordRepository.java
@@ -89,6 +89,8 @@ public class RecordRepository extends AbstractRepository {
 
         consumer.close();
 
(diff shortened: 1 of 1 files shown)


tchiotludo:
Good catch :smile: 
Already have the same feeling looking at some topic. 

Sorry for the delay: :beach_umbrella: and busy with last release ! 

Thanks :+1:

Issue #336, opened 2020-07-17
match group with ldap


cpoule23:
Hi all,

I try to use ldap authentification with group binding (username : cpoule23) but it's not working :/

My config :

    configuration: |
      akhq:
        server:
          access-log: 
            enabled: true
            name: org.akhq.log.access
        security:
          default-group: no-roles
          groups:
            topic-reader:
              name: "topic-reader"
              roles:
                - topic/read
              attributes:
                - topics-filter-regex: "*"
            topic-writer: # just a key, no matter will be override by name below
              name: "topic-writer" # Group name
              roles:
                - topic/read
                - topic/insert
                - topic/delete
                - topic/config/update
              attributes:
                topics-filter-regexp: "*"
            admin2: # just a key, no matter will be override by name below
              name: "admin2" # Group name
              roles:
                - topic/read
                - topic/insert
                - topic/config/update
                - topic/config/update
                - topic/delete
                - topic/config/update
                - node/read
                - node/config/update
                - topic/data/read
                - topic/data/insert
                - topic/data/delete
                - group/read
                - group/delete
                - group/offsets/update
                - acls/read
                - registry/read
                - registry/insert
                - registry/update
                - registry/delete
                - registry/version/delete
                - connect/read
                - connect/insert
                - connect/update
                - connect/delete
                - connect/state/update
          ldap:
            group:
              CS-MOI_stat:
                groups:
                  - admin2
      micronaut:
        security:
          ldap:
            default:
              enabled: true
              context:
                server: 'ldap://myldap:389'
                managerDn: 'uid=akhq_ibuy,ou=manager,o=mycompanie'
                managerPassword: 'mypass'
              search:
                base: "o=mycompanie"
                attributes: "uid"
              groups:
                enabled: true
                base: "ou=distributionlist, ou=application, ou=group, o=mycompanie"
                attribute: "cn"
                filter: "(&(objectclass=ocStaticGroup)(cn=CS-*)(uniqueMember=uid=cpoule23*))"

In debug, I get :
```
2020-07-17 09:17:32,018 DEBUG pGroup-1-2 LdapAuthenticationProvider Starting authentication with configuration [default]
2020-07-17 09:17:32,018 DEBUG pGroup-1-2 LdapAuthenticationProvider Attempting to initialize manager context
2020-07-17 09:17:32,019 DEBUG pGroup-1-2 LdapAuthenticationProvider Manager context initialized successfully
[...]

PR #298, opened 2020-05-29
WIP: New React UI

polarising-java:
Added Live Tails
Remove duplicate CSS
Fix bugs

Issue #161, opened 2019-12-03
Who's using KafkaHQ ! Help wanted !


tchiotludo:
Since KafkaHQ is an open source project, it's really complicated to gather usage on this project.
I know that the project is used by a lots of company but it's really hard to have a complete overview.

So please submit a [Pull Request](https://github.com/tchiotludo/kafkahq/edit/dev/README.md) on the section if you want to add publicly that your company is using KafkaHQ : https://github.com/tchiotludo/kafkahq#whos-using-kafkahq.

If you have some details on usage (size of installation, kind of usage, ....), please feel free to add a description that will help me to prioritize the next feature.

Issue #142, opened 2019-11-07
Oauth2 Authentification 


tchiotludo:
Since micronaut support [oauth2](https://micronaut-projects.github.io/micronaut-security/latest/guide/#oauth2-authorization-code), it will be fine to have this support on KafkaHQ

zthulj:
I've got a question.
Currently, to set the oauth2 config in micronaut, it's required to use 
```
micronaut:
  security:
    oauth2:
      ...
```

To be more consistent, I thought we could set the oauth2 settings under 
```
kafkahq:
  security:
    oauth2:
      ...
```

I've discussed a bit with the community about this. To do that, we just need to @Replace the oauth configuration bean, to make sure it take config from the key we want.
Do you want to do this or to let the default micronaut key ?

tchiotludo:
to be honest, I just can't think you could change the `micronaut.securty.oauth` by a custom one. 

I imagine It can be done with an extends to the base config class, just change the configuration prefix, and add @Replaces. no ? 

To be honest, I'm not really sur to change this one. (not really sure, just first impression). 
The advantage with keep micronaut one,  just drop a link in kafkahq docs to micronaut docs and documentation is done and up to date. 
In the others case, you will need to keep sync the documentation, and since it not my preferred flavor :smile: , I'm not really sure to want that. 

seems to be simpler like that, no ?

zthulj:
Yes, exactly how you have to do it apparently, but I haven't tried it yet.
I understand the sync documentation argument, and, I rather agree with that. It's not my preferred flavor either 😆.
 
However, this are the two points I had in mind when asking this : 

- link to the micronaut documentation will lead for the user to think he has something to do with 
an implementation of OauthUserDetailsMapper, because it's one of the pre-requisites in the documentation. 
However, to avoid this, we could just add a little bit of informations in an oauth2 section in the documentation.

- Speaking of OauthUserDetailsMapper impl, we will provide some, compatible with the different oauth2 providers (keycloak, google, amazon..).
We'll have to create one per provider, or try to be generic, I guess, because the detail of the users can be different from provider to provider. (I am not 100% sure about this, I have to check the result of getUserDetail from at least 2 providers)

The workflow , as I understand it, is the following:

1. Redirect to provider login page
2. Login
3. Callback to kafkaHQ with an access token
4. Call the provider api to get user details using the access token.
5. Convert user details from the provider to UserDetails of micronaut.

4 and 5 are the userDetailsMapper's job. 

[...]

PR #342, opened 2020-07-27
Add OIDC login

jniebuhr:
This PR adds OIDC capability to AKHQ. 
Username and roles will be taken from the OpenID claims and translated using the existing logic.
When activating OIDC login, the login page will show a redirect button instead of the login form.


jniebuhr:
Probably solves #142

tchiotludo:
Wohh !! 
Thanks for this one :+1: really appreciate ! 

Just some thinking : 
The most important : This PR will break this [one](https://github.com/tchiotludo/akhq/pull/298). 
I will change the whole ui with a new react ui. 
The main idea for now is to release a new version with both ui (old and new), this must be done this week I think (in order to gather bug).

The next version (after the one with both) will remove the old ui in favor in the react one. 
In this release, I will remove the session in favor of a JWT cookie (that will be compatible with your PR).
I think you will need to wait a little bit for this release in order to be merged. 

Other concern about the PR : 
- I want to be able to have both LDAP / basic auth with Oauth, so we won't disabled the login form if there is an auth 
- I want to be able to support multiple Oauth provider, so maybe `Oidc` must be `List<Oidc>`

jniebuhr:
When the React UI is merged, I guess we'll need some configuration endpoint that returns which providers are enabled then.
For your other two concerns, I'll adjust the PR.

jniebuhr:
Screenshot from 2020-07-27 11-10-34

@tchiotludo 
I've changed it so you can configure multiple oidc providers and it doesn't hide the normal login if you do.

jniebuhr:
I think the check is broken? It's been at the same step for quite some time.

jniebuhr:
@tchiotludo 
It looks like you merge some stuff in for the new UI. I've resolved the conflicts for this PR. 
Should I also add the OIDC logic to the React UI? If yes, do you want to pass the configuration options to the frontend with an API endpoint?

tchiotludo:
Yes the old ui will be [remove](https://github.com/tchiotludo/akhq/pull/348) on next release ! 

Here you can find the [endpoint](https://github.com/tchiotludo/akhq/blob/3cd3c81d7f527830dd87f57f683581d4b1558293/src/main/java/org/akhq/controllers/AkhqController.java#L66-L76) where you can add the OpenId Providers

jniebuhr:
@tchiotludo there you go, implemented in react now

JacobJohansen:
I think Micronauts may need to be bumped to 1.3.7 to resolve an issue with host resolver https://github.com/micronaut-projects/micronaut-core/issues/3416

specifically when for the use-case when the proxy handles SSL termination

JacobJohansen:
> I think Micronauts may need to be bumped to 1.3.7 to resolve an issue with host resolver [micronaut-projects/micronaut-core#3416](https://github.com/micronaut-projects/micronaut-core/issues/3416)
> 
> specifically when for the use-case when the proxy handles SSL termination



> I think Micronauts may need to be bumped to 1.3.7 to resolve an issue with host resolver [micronaut-projects/micronaut-core#3416](https://github.com/micronaut-projects/micronaut-core/issues/3416)
> 
> specifically when for the use-case when the proxy handles SSL termination

scratch that, you have to enable forwarding headers; the documentation made it seem as if Micronauts did that by default

tchiotludo:
I confirm it's enabled by default.
But there is case you need to fine tune it, for example behind a GCP LB : 
```
    micronaut:
      server:
        host-resolution:
          host-header: Host
          protocol-header: X-Forwarded-Proto
```
Because micronaut expect both `X-Forwarded-Proto` & `X-Forwarded-Host` and Google LB don't send the `X-Forwarded-Host`

src/main/java/org/akhq/configs/Oidc.java
```diff
@@ -0,0 +1,26 @@
+package org.akhq.configs;
+
+import io.micronaut.context.annotation.ConfigurationProperties;
+import io.micronaut.security.oauth2.endpoint.token.response.OpenIdClaims;
+import lombok.Data;
+
+import java.util.Map;
+
+@ConfigurationProperties("akhq.security.oidc")
```
JacobJohansen, line 9: should this be a configuration each?

tchiotludo: yes I think it must be `@EachProperty`

jniebuhr: Nope, it is not an array so it has to be `@ConfigurationProperties`

tchiotludo: my fault ;) 
To be honest, I don't have time to look deeply ;)

src/main/java/org/akhq/modules/OidcUserDetailsMapper.java
```diff
@@ -0,0 +1,115 @@
...
+     * @param providerName The OpenID provider name
+     * @param tokenResponse The token response
+     * @param openIdClaims The OpenID claims
+     * @return The attributes to set in the {@link UserDetails}
+     */
+    @Override
+    protected Map<String, Object> buildAttributes(String providerName, OpenIdTokenResponse tokenResponse, OpenIdClaims openIdClaims) {
+        Map<String, Object> attributes = super.buildAttributes(providerName, tokenResponse, openIdClaims);
+        userGroupUtils.getUserAttributes(getOidcRoles(providerName, openIdClaims)).forEach(attributes::put);
+        attributes.remove("roles");
```
JacobJohansen, line 89: should this be the roles field?

JacobJohansen: I was using the groups claim from Azur AD and It output to many AD/Office 365 groups for the cookie to store in the browser. Dropping the roles field cleaned up that space.

jniebuhr: I‘m intentionally dropping a field called „roles“ because that is where Micronaut inserts the resolved roles. It will not do that if there is an attribute called „roles“.

For the other problem, rather than just copying everything over from the provider JWT we should probably add a whitelist.

tchiotludo: To be honest, it's this part that hurt me, storing the roles in the JWT cookie.

In a future release, I will need to use another logic than role for akhq, people want to have role per cluster, that lead to incompatibility with the micronaut role.

I would have prefer to have only username in the JWT token, and server will add the role on the login part.

This will allow to change the role on server side without any impact on the cookie part and we won't have anymore the cookie size issue.

@jniebuhr what do you think about that ?

jniebuhr: If we just Store the username in the cookie there won’t be a way to get the roles for the user, other than caching it in-memory or doing the OIDC flow all over again. We could store all the unresolved roles that had matches in the cookie, that should be smaller.

tchiotludo: In another project, I use this to handle custom mapping : 
```java
import io.micronaut.security.oauth2.endpoint.token.response.DefaultOpenIdUserDetailsMapper;

@Singleton
@Replaces(DefaultOpenIdUserDetailsMapper.class)
public class GlobalOpenIdUserDetailsMapper extends DefaultOpenIdUserDetailsMapper {

    public GlobalOpenIdUserDetailsMapper(OpenIdAdditionalClaimsConfiguration openIdAdditionalClaimsConfiguration) {
        super(openIdAdditionalClaimsConfiguration);
    }

    @Override
    @Nonnull
    public UserDetails createUserDetails(String providerName, OpenIdTokenResponse tokenResponse, OpenIdClaims openIdClaims) {
        // user is here : openIdClaims.getEmail();
        // mapping username using configuration files 
        //  return new UserDetails();
    }
]
```
It works very well and nothing except the user is store on the cookie. 
The oauth flow is only done 1 time also.

jniebuhr: Do you have a username to roles mapping in the Config then?

tchiotludo: Yes but in this project I have a database to fetch the roles.

That lead to a question, where is the role in this actual implementation ?

jniebuhr: The roles get passed in the provider token as a claim, I then map them to AKHQ roles. Those get stored in the JWT and can be accessed on every request.

tchiotludo: So you have some openid provider where you can customize the claim content ? (add specific fields for an application) ?  

Most of users will use a generic openId provider (Google, Microsoft, ...) and will not be able to customize the claims (IMO)

jniebuhr: You can’t always customize which claims are there but you can usually select them well enough. If the provider has no understanding about roles you’ll need to use a default role anyway. In our case we have full control over claims, yes.

jniebuhr: Maybe we can add a default role for OIDC authenticated users

JacobJohansen: what I have done in spring before but I believe it would be equivalent to replace the bean UserDetails and override getRoles to map the group "role" form the IDP to the RBAC roles that have been defined in the configuration. It will help keep the JWT token smaller. https://github.com/micronaut-projects/micronaut-security/blob/39eef7d3d3cfe6a25ba957026d725a78b769cbb8/security/src/main/java/io/micronaut/security/authentication/UserDetails.java

JacobJohansen: > Most of users will use a generic openId provider (Google, Microsoft, ...) and will not be able to customize the claims (IMO)

to support this use-case there will probably need to be username to group "role" mapping in the future and then the UserDetails mapper can output the roles that are referenced on the endpoint annotation

tchiotludo: Agree with that, but I also think it's a must have for OpenId.
Most of users will use a public / not customizable providers so we need to allow user to have a default role but also have a special mapping for username > roles for admin for example :) 
I will hope to have a better mapping (group based) but I don't see a better way for now :cry:

jniebuhr:
@tchiotludo @JacobJohansen I have reworked the whole users and groups mapping for OIDC to work like the LDAP one. I've also added a default group to both so they can share their code. I've renamed the ldap.user and ldap.group properties to ldap.users and ldap.groups but kept the old ones around for people to migrate.

tchiotludo:
Starting to look at deeply, seems to be a really good stuff here ! 
But I'll try to make it work with google oauth, I don't find for now. 
The redirect to callback is ok, but the custom mapper is not called, so the user has no role and will be redirect to login page. 

Don't understand for now :thinking:

tchiotludo:
To make it detect Mapper by micronaut, I must changed to this  : https://github.com/tchiotludo/akhq/commit/29f40564bc25685cf7d50904409c3ef6f32f6504

Maybe Google provider have not the same behaviour than yours ? 

I also changed the `GroupMapping` & `UserMapping` to a `List` since it don't work (think must be `@EachProperties` to work with a map (but I don't want to use Map in configuration files, since micronaut change the case of the key, it lead to a lot of bug like that https://github.com/tchiotludo/akhq/issues/336#issuecomment-663367630)

tchiotludo:
In a meantime, I've merged this one : 
https://github.com/tchiotludo/akhq/pull/380

Seems there is changed on the OpenId on the 2.0.0.

Also maybe, we should disabled session login to JWT everywhere ! AKHQ will work behind a load balancer without sticky session !
My only concern is `secret` (provide a default one or not). 

Maybe we must start to enable the JWT first to avoid a biggest PR ?

jniebuhr:
I'll check the whole implementation based on micronaut 2 and do a test with Google. I'll update the PR afterwards

jniebuhr:
> To make it detect Mapper by micronaut, I must changed to this : [29f4056](https://github.com/tchiotludo/akhq/commit/29f40564bc25685cf7d50904409c3ef6f32f6504)
> 
> Maybe Google provider have not the same behaviour than yours ?
> 
> I also changed the `GroupMapping` & `UserMapping` to a `List` since it don't work (think must be `@EachProperties` to work with a map (but I don't want to use Map in configuration files, since micronaut change the case of the key, it lead to a lot of bug like that [#336 (comment)](https://github.com/tchiotludo/akhq/issues/336#issuecomment-663367630))

If we change the Map to a List, it's gonna be a different syntax to the LDAP one

jniebuhr:
@tchiotludo I have changed the config to always use JWT now. I've tested the code with the pasted config, it works just fine using basic auth and oidc at the same time. It'll generate JWT tokens from the user details and store them in a cookie. A user upgrading to this will need to set a secret. For security reasons, we shouldn't provide a default one. If you want I can extract the JWT part from this PR, but it isn't a lot to make it work. Mostly adding the dependency and some config.

```
akhq:
  connections:
    plain:
      properties:
        bootstrap.servers: "kafka:9092"
      schema-registry:
        url: "http://schema-registry:8085"
      connect:
      - url: "http://connect:8083"
  security:
    default-group: no-roles
    basic-auth:
      demo:
        password: 2a97516c354b68848cdbd8f54a226a0a55b21ed138e207ad6c5cbb9c00aa5aea # demo
        groups:
          - admin
    oidc:
      enabled: true
      providers:
        google:
          label: "Login with Google"
          username-field: email
          groups-field: roles
          default-group: no-roles
          users:
            some-admin:
              username: <your gmail address here>
              groups:
                - admin

micronaut:
  server:
    port: 8080
  security:
    enabled: true
    oauth2:
      enabled: true
      clients:
        google:
          client-id: <client-id>
          client-secret: <client-secret>
          openid:
            issuer: "https://accounts.google.com"
    token:
      jwt:
        signatures:
          secret:
            generator:
              secret: thisisarandom256bitsecretyoullneedtochange

```

src/main/resources/application.yml
```diff
@@ -20,13 +20,18 @@ micronaut:
         mapping: "/swagger/**"
   security:
     enabled: true
```
tchiotludo, line 22: Maybe we need to disabled security by default, since there is not `secret` defined and add some doc on readme about that (enabled security & add a secret)

src/main/java/org/akhq/configs/Ldap.java
```diff
@@ -0,0 +1,56 @@
...
+@ConfigurationProperties("akhq.security.ldap")
+@Data
+public class Ldap {
+    private String defaultGroup;
+    private Map<String, GroupMapping> groups = new HashMap<>();
+    private Map<String, UserMapping> users = new HashMap<>();
+    /**
+     * @deprecated Naming isn't consistent with the other mappings
+     */
+    @Deprecated(forRemoval = true)
```
tchiotludo, line 23: Go BC for a good time ! remove it ;) as `user`

src/main/java/org/akhq/configs/Ldap.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@ConfigurationProperties("akhq.security.ldap")
+@Data
+public class Ldap {
+    private String defaultGroup;
+    private Map<String, GroupMapping> groups = new HashMap<>();
```
tchiotludo, line 18: There is a lot of issue using a Map<String, T>, micronaut change the case of key and user complain about that.
I think we must go with a List<T> here  as `group`

tchiotludo (changes requested):
Added some minor change ! 
I would prefer to go breaking compatibility for the whole next release since we change the session to cookie, there is breaking change. (also remove the old ui) 
So I think we must go to a full breaking change !

src/main/resources/application.yml
jniebuhr: Yes, that's a good idea i guess

src/main/java/org/akhq/configs/Ldap.java
jniebuhr: Alright, if we go breaking that's no problem :+1:

src/main/java/org/akhq/configs/Ldap.java
jniebuhr: Yeah, I only added it this way to stay consistent with the other entries. If we change it here, I'll change it in all occurences (oidc, ldap, basic auth, general groups)

jniebuhr: There is a weird bug with `@EachProperty` in list mode. If you use it, it will not fill List<String> fields below it. I had to remove any `@EachProperty` and use a new `SecurityProperties` class for it.

tchiotludo:
I just try, it work perfectly ! :heart: 
The release must contains a big warning for all the breaking change :disappointed_relieved: but was necessary I think ! 
No more issue on this. 

Really thanks for the hard work and let people try it on dev branch before the next release ! 

:+1:

tchiotludo:
seeing your email, maybe you can contribute here also : https://github.com/tchiotludo/akhq/issues/161 :smile:

PR #352, opened 2020-08-07
GenericRecord to JSON serialization with correct decimal output

rigadon:
Bugfix of GenericRecord to JSON serialization for decimals.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -3469,6 +3469,15 @@
       "resolved": "https://registry.npmjs.org/binary-extensions/-/binary-extensions-2.1.0.tgz",
       "integrity": "sha512-1Yj8h9Q+QDF5FzhMs/c9+6UntbD5MkRfRwac8DoEm9ZfUBZ7tZ55YcGVAzEe4bXsdQHEk+s9S5wsOKVdZrw0tQ=="
     },
+    "bindings": {
+      "version": "1.5.0",
+      "resolved": "https://registry.npmjs.org/bindings/-/bindings-1.5.0.tgz",
+      "integrity": "sha512-p2q/t/mhvuOj/UeLlV6566GD/guowlr0hHxClI0W9m7MWYkL1F0hLo+0Aexs9HSPCtR1SXQ0TD3MMKrXZajbiQ==",
+      "optional": true,
+      "requires": {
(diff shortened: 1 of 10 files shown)


tchiotludo:
Seems to be strange bug fix ! 
Can you tell me more ? 

The only goal of this PR is to have the right number of decimal right ? 
Before it was displayed as float ?

rigadon:
The problem here is that we are working with GenericRecord and it is unable to serialize decimals correctly because inside of the record we have an array of bytes. It causes an issue with JSON we display. For instance, in you look at https://github.com/tchiotludo/akhq/blob/3d3b2d67bb4077e3cefdc43fff15834f8543c763/src/test/java/org/akhq/modules/AvroToJsonSerializerTest.java and try to serialize a generic record that contains Dog avro to string via `toString` method then you will receive a string like `{"id": 10, "name": "Tiger", "weight": "\u0004\u0010"}`. At the same time with the change, we will be able to detect those decimal values based on schema an serialize that record to `"{\"id\":10,\"name\":\"Tiger\",\"weight\":\"10.40\"}"`.

tchiotludo:
Really strange ! I'll try your PR !
without your PR : 
image

With your PR : 
image

I have better result without ! 
To test it, i simply produce a message with Dog schema.

The decimal part (max: 2) is not respected but the value is correct. 
Your PR seems to display the wrong value.

Maybe I lost a thing ?

rigadon:
Unfortunately, you faced another defect AKHQ has. It looks like you submitted a message via the interface AKHQ has and the decimal was provided as a string. In this case, in the current implementation, the string("10.40") will be serialised to an array of bytes and deserialised from the array of bytes to the string and it looks like the value is correct in Kafka but in fact, if you deserialise that array of bytes to decimal you will receive a completely different number(2112617318.88). Basically, that happened with you and I understand that it is a bit confusing. I did a test when I produced a Dog avro message to a topic via another application and tried to read that message with two different implementations of AKHQ.

With the current implementation I got:
image

And with the fix I made I got:
image

Of course, we also need to fix an issue you have faced but I expect it to be fixed in another PR.

tchiotludo:
OK make sense ;) 
Thanks for the PR :) 

I add an issue for the produce, feel free to take it ;)

PR #373, opened 2020-08-26
Migrate topic list config button to new ui

Gagarmel:
A configuration button has been added to topiclist in #324 that was missing in the new UI.
So i've migrated the new button into the new UI.


tchiotludo:
God catch ! 
Thanks ;)

PR #376, opened 2020-08-28
Fix error selecting connector properties

polarising-java:
Fix Create button position
Fix select dropdown size


tchiotludo:
perfect :+1:

Issue #332, opened 2020-07-10
Add share button


tchiotludo:
A share button for every message and add a route to show only this message

polarising-java:
@tchiotludo  create a new API for this?

   @Secured(Role.ROLE_TOPIC_DATA_READ)
    @Get("api/{cluster}/topic/{topicName}/data/record")
    @Operation(tags = {"topic data"}, summary = "Ge a single record")
    public Record record(
        HttpRequest<?> request,
        String cluster,
        String topicName,
        Optional<Integer> partition,
        Optional<Integer> offset
    )

tchiotludo:
in think this is the better way, yeah :) 
Since we need to be sure that only that exactly offset is fetch, I think we need a new route for that :+1:

tchiotludo:
done in #401

PR #401, opened 2020-09-18
WIP addsharebutton

polarising-java:
Add share button #332 
Improve get messages function 

@tchiotludo i added a icon to copy to clipboard:
image


client/src/containers/Topic/Topic/TopicData/TopicData.jsx
```diff
@@ -230,6 +242,18 @@ class TopicData extends React.Component {
     });
   }
 
+  async handleOnShare(row) {
```
tchiotludo, line 245: I think I mostly also want to go the page to see the share result ?

src/main/java/org/akhq/controllers/TopicController.java
```diff
@@ -320,6 +320,30 @@ public Record deleteRecordApi(String cluster, String topicName, Integer partitio
             });
     }
 
+    @Secured(Role.ROLE_TOPIC_DATA_READ)
+    @Get("api/{cluster}/topic/{topicName}/data/record")
```
tchiotludo, line 324:
```suggestion
    @Get("api/{cluster}/topic/{topicName}/data/record/{partition}/{offset}")
```

src/main/java/org/akhq/controllers/TopicController.java
```diff
@@ -320,6 +320,30 @@ public Record deleteRecordApi(String cluster, String topicName, Integer partitio
...
+    @Operation(tags = {"topic data"}, summary = "Get a single record by partition and offset")
+    public ResultNextList<Record> record(
```
tchiotludo, line 326: I would go to path parameter instead of querystring since the argument are mandatory for offset and partition
and change 
```java
Optional<String> after,
Optional<Integer> partition
```
to 
```java
String offset,
Integer partition
```

src/main/java/org/akhq/repositories/RecordRepository.java
```diff
@@ -140,6 +140,25 @@
         }, "Offsets for " + partitions + " Timestamp " + timestamp, null);
     }
 
+    public Optional<Record> consumeSingleRecord(String clusterId, Topic topic, Options options) throws ExecutionException, InterruptedException {
+        return Debug.call(() -> {
+            Optional<Record> singleRecord = Optional.empty();
```
tchiotludo, line 145: maybe tune the option `max.poll.records` to 1 to avoid to poll too many message ?

tchiotludo (changes requested):

src/main/java/org/akhq/repositories/RecordRepository.java
polarising-java: yes, missing that.

src/main/java/org/akhq/controllers/TopicController.java
polarising-java: Maybe 
```
Integer offset,
Integer partition
```

We only pass the offset number, and do the join for dataSearchOptions after. 

In RecordRepository.getFirstOffsetForSortOldest the 
`first = options.after.get(partition.getId()) + 1;`  
will never show offset 0

client/src/containers/Topic/Topic/TopicData/TopicData.jsx
polarising-java: open in a new window?

tchiotludo: Just go to the share page in the same window it will be fine

tchiotludo:
I have a last commit to make it works well ! 
Just the last merge is not so easy, can you do it please ?

polarising-java:
I did the merge with DEV.

The change in 
` first = options.after.get(partition.getId()) + 1;` 
to
` first = options.after.get(partition.getId()) ;` 
is ok?

tchiotludo:
Yes seems to be ok on my test, thanks 👍

PR #402, opened 2020-09-18
Allow security configuration without basic auth

dorsradek:
Currently, we need to specify basic auth section in `akhq.security` even if we don't use it.

If not NullPointerException is thrown in https://github.com/tchiotludo/akhq/blob/cbcaf627216a1c3650323e0611a9f95013c25779/src/main/java/org/akhq/controllers/AkhqController.java#L69


tchiotludo:
Thanks :+1:

Issue #397, opened 2020-09-14
Handle a loading state on all table


tchiotludo:
For now, during initial load, the warning message is displayed `No data available`.

It will be nice to have a loading message instead.

PR #404, opened 2020-09-18
Handle a loading state on all table 

polarising-java:
Handle a loading state on all table  #397


tchiotludo:
Perfect !! 
Just change with a spinner instead of alert :smile: 
Thanks :+1:

PR #408, opened 2020-09-23
Feat/304 support references

mredjem:
Manage Avro schema references coming with Confluent schema registry 5.5.0:

- manage Avro schema references to read messages from topics
- manage Avro schema references to display schemas from the schema registry
- add possibility to register a schema with its type references in the UI (old way with only a json avro schema still compatible)

Register a schema with references:
```json
{
  "schema": "...",
  "references": [
    {
      "name": "Schema1",
      "subject": "schema_1",
      "version": 1
    }
  ]
}
```

Schema registry now supports Protobuf and JSON, this change targets only Avro for now.


src/test/java/org/akhq/repositories/SchemaRegistryRepositoryTest.java
```diff
@@ -55,11 +56,19 @@
         .name("meta").type().nullable().map().values().stringType().noDefault()
         .endRecord();
 
+    public final static String SUBJECT_3 = "SCHEMA_3";
+    public final static String SCHEMA_3 = "{\"type\":\"record\",\"name\":\"Schema3\",\"namespace\":\"org.akhq\",\"fields\":[{\"name\":\"name\",\"type\":[\"null\",\"string\"]}]}";
+
+    public final static String SUBJECT_4 = "SCHEMA_3";
```
tchiotludo, line 62:
```suggestion
    public final static String SUBJECT_4 = "SCHEMA_4";
```

tchiotludo (changes requested):
Seems to do the jobs (I think, I don't use reference for now).
Just seems strange that test work with wrong subject ?

tchiotludo:
also seems to break the old way to register schema : 

image

I only put the `SCHEMA_3` and I was unable to register the schema

mredjem:
Fixed issue with the schema registering and the unit test.

tchiotludo:
Seems to be work now !! 

Thanks :+1:

PR #409, opened 2020-09-24
Produce big decimals

niels-zopa:
This introduces `BigDecimalFriendlySpecificDatumReader` so that when users provide a JSON message that includes decimals, AKHQ will correctly write those decimals to Kafka.

This is sort of a converse to https://github.com/tchiotludo/akhq/pull/352: that PR displays BigDecimals as strings in the output, and this allows you to use the same syntax to input BigDecimals.


tchiotludo (changes requested):
don't really like the naming here ! 
`BigDecimalFriendlySpecificDatumReader` must be more generic since it will be a `GenericDatumReader` for the whole app (not only BigDecimal for also other special logicalType)

niels-zopa:
I wanted to keep the naming similar to the `BigDecimalFriendlySpecificDatumWriter` introduced here: https://github.com/tchiotludo/akhq/pull/352/files#diff-cb15c974c793f37d548435a3349189ccR28

So I think if we're going to change the `BigDecimalFriendlySpecificDatumReader`, we should also change the `BigDecimalFriendlySpecificDatumWriter`.

In what way do you think it should be more generic? I don't think there are any other logical types that we have to read in a special way, are there?

tchiotludo:
Good point, renaming the other to be consistent is the good way to handle this. 
For other, I don't really know for now, but to be honest, I was thinking that BigDecimal will not be a trouble, maybe other one will have the issue.

niels-zopa:
I've added tests for another logical type (date). The BigDecimalFriendly readers and writers don't in fact do anything special with any of the other logical types --- the other logical types are treated in exactly the same way as before this PR.

tchiotludo:
I think I don't explain very well what I mean :smile: 
I mostly think to rename `BigDecimalFriendlyGenericDatumReader` to something like `ConverterGenericDatumReader`. 
For now, like you add on the test, the date are displayed as timestamp, I want in a near future to convert to ISO datetime (if you have time :smile:)
Since it will be the same Reader that will handle it, we need to have a more generic name ? 

Do you see what I mean ?

niels-zopa:
Yes, thanks for clarifying. I see what you mean: in the future, it should be possible to output and input other logical (such as dates or timestamp) in a human-readable way as well.

I agree that it would be nice to display timestamps as dates. Actually, I already spent quite some time looking into that and tried many things, but without success. The approach we're taking here for BigDecimals is very hacky: the `decimal` logical type has bytes as its underlying simple type, and we simply change the bytes to be the byte sequence for a nice string. However, the same approach doesn't work for timestamps or dates: the date logical type (see [this part of the avro spec](https://avro.apache.org/docs/current/spec.html#Date)) is an `int`, and it is impossible (as far as I know) to convince the reader/writer to accept anything else than an `int`. (I tried many different things, and at some point you run into some issues very in very low-level code in the Avro code base.) As a consequence, if the schema expects a timestamp, we have to print some kind of `int`.

Anyway, I don't mind changing the name. What do you think of `HumanFriendlyGenericDatumReader` and `HumanFriendlyGenericDatumWriter`?

niels-zopa:
By the way, if you really want to also pretty-print other logical types such as dates, I think there are two solutions (but they are both quite a lot of work):

1. Change the schemas: if you need to write a schema that has a timestamp somewhere, create a new schema that has a string in that place, and then write the data using that schema. Something similar should happen in the other direction.
2. Create custom serialisers and deserialisers (from `GenericRecord` to JSON).

Those are the solutions I came up with -- maybe there are more/better solutions.

niels-zopa:
Hey @tchiotludo , is there anything more I can do to help you with this PR?

tchiotludo:
To be honest, I need to find some time to look at this one more deeply. 
I've some topic BigDecimal and with the last veresion there are not displayed correctly : 
image

I suspect we are going the wrong at first on deserialize, so need to be sure of that before getting on serialization.
If the deserizalization don't work, we have a bigger issue than I think

niels-zopa:
Ok, thanks for the quick reply! Just let me know if I can help.

magnusgundersen:
Any progress on this? The feature would be much appreciated.

PederAaby:
Our team would welcome this feature very much as well :)

tchiotludo:
hi there :) 
I have made a complete rework about that on `dev`, all logical type must be supported !
can you try ?

PR #411, opened 2020-09-24
Fix LDAP example in docs.

AntonSmolkov:
Fix according to this issue and comment
https://github.com/tchiotludo/akhq/issues/336#issuecomment-663367630

We have spent hours struggling with ldap groups containig dots.
With this (ugly, i know) approach no one will ever face this tricky problem.

Also, example from documetation with testing server does not work at all for now, this PR fixes this too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,8 +359,9 @@ Define groups with specific roles for your users
 
 ##### LDAP
 Configure how the ldap groups will be matched in AKHQ groups 
-* `akhq.security.ldap.groups`: Ldap groups list
-  * `- name: ldap-group-name`: Ldap group name (same name as in ldap)
+* `akhq.security.ldap.group`: Ldap group yaml section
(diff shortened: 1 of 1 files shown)


tchiotludo:
Hi, 
What version of akhq do you use ? 

One wrong thing in akhq is `README.md` is based on `dev` branch and we have changed the way all of this works for next release with BC changed. 

Seems that your main issue here !

AntonSmolkov:
Latest image from DockerHub. Guess it's 0.15.0

AntonSmolkov:
0.15.0 tag still contains ldap example which does not work with non-lowercase ldap-gropus

tchiotludo:
0.16.0 change the configuration and this PR will not work. 
Maybe you can update the PR with last version, there is some interesting MS ActiveDirectory comment I think

tchiotludo:
closing for inactivity

PR #424, opened 2020-09-30
fea(ui): allow to select topics in consumer group when update offsets

### LINKED ISSUE #339: Filter topics in consumer group when update offsets
Hi, we have consumer groups with many topics attached and if we use update offsets, this function change offset on all topic in consumer group. If we want to update only one topic, we have to select the desired change (for example, timestamp), copy the offset across all partitions of the desired topic, then go back and insert them manually. If it possible to add selection of topics for update offsets function?

tchiotludo:
close #339

### DIFF
--- a/client/src/containers/ConsumerGroup/ConsumerGroupDetail/ConsumerGroupUpdate/ConsumerGroupUpdate.jsx
+++ b/client/src/containers/ConsumerGroup/ConsumerGroupDetail/ConsumerGroupUpdate/ConsumerGroupUpdate.jsx
@@ -124,14 +124,11 @@ class ConsumerGroupUpdate extends Form {
     this.setState({ formData });
   };
 
-  unCheckAll = ()  => {
+  unCheckAll = (value)  => {
     const {checked} = this.state;
-    let toCheck;
 
     Object.keys(checked).forEach(name => {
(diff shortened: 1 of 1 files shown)


client/src/containers/ConsumerGroup/ConsumerGroupDetail/ConsumerGroupUpdate/ConsumerGroupUpdate.jsx
```diff
@@ -133,60 +148,71 @@ class ConsumerGroupUpdate extends Form {
...
-        <fieldset id={`fieldset-${topidId}`} key={topidId}>
-          <legend id={`legend-${topidId}`}>{topidId}</legend>
-          {this.renderPartitionInputs(groupedTopicOffset[topidId], topidId)}
+        <fieldset id={`fieldset-${topicId}`} key={topicId}>
+          <legend id={`legend-${topicId}`}>
+            <input
+              type="checkbox"
+              value={topicId}
+              defaultChecked={checked[topicId]}
+              onChange={this.checkedGroupedTopicOffset}/> {topicId}
```
tchiotludo, line 198: @polarising-java : need some help here please :pray: 
I need to click several time on the checkbox to make it work ! 
don't understand, the uncheck all button works fine !

PR #425, opened 2020-09-30
Support BCrypt password encryption

gquintana:
SHA-256 without salt is considered as breakable.

This pull request adds:
* Crypt format support https://en.wikipedia.org/wiki/Crypt_(C)
* BCrypt password hashing relying on https://www.mindrot.org/projects/jBCrypt/

However it still doesn't support SHA-256 with salt and multiple iterations as https://passlib.readthedocs.io/en/stable/lib/passlib.hash.sha256_crypt.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,7 +348,9 @@ Define groups with specific roles for your users
 ##### Basic Auth
 * `akhq.security.basic-auth`: List user & password with affected roles 
   * `- username: actual-username`: Login of the current user as a yaml key (may be anything email, login, ...)
-    * `password`: Password in sha256, can be converted with command `echo -n "password" | sha256sum`
+    * `password`: Password in sha256 or bcrypt. The password can be converted 
+      * For sha256, with command `echo -n "password" | sha256sum` or Ansible filter `{{ 'password' | hash('sha256') }}`
+      * For BCrypt, with Ansible filter `{{ 'password' | password_hash('blowfish') }}`
     * `groups`: Groups for current user
 
 > Take care that basic auth will use session store in server **memory**. If your instance is behind a reverse proxy or a
--- a/README.md
+++ b/README.md
@@ -348,16 +348,28 @@ Define groups with specific roles for your users
 ##### Basic Auth
 * `akhq.security.basic-auth`: List user & password with affected roles 
   * `- username: actual-username`: Login of the current user as a yaml key (may be anything email, login, ...)
-    * `password`: Password in sha256 or bcrypt. The password can be converted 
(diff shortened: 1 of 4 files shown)


tchiotludo:
I don't consider this for now.
Why not, but I would prefer to have a configuration option to change the crypt for password and all the user will use the same encryption method. 
Also can you add some information on README.md about that ?

gquintana:
BCrypt requires Crypt format ($2?$.*), You may notice I don't remove the prefix with BCrypt.
BCrypt is a modern password hashing algorithm supported by many tools: Ansible in my case.
I would prefer a JVM supported algorithm (like sha256), sadly BCrypt, Argon and PBKDF2 are the accepted password hashing algorithms.
I can remove all the stuff to handle $1$, $5$ and $6$ prefix support if you wish.
I'll fix readme and failing checks.

tchiotludo:
It's why I'm talking about a configuration property that will tell if the user want bcrypt or sha.
It will remove the receipt and if else to try to find which algorithm is used

gquintana:
I truly understand that this prefix parsing sucks.
However you can not remove $2$ prefx because this is not how BCrypt works.
The configuration property should tell if it's crypt format or not. Not the algorithm id.
Also note that Crypt format and Python's Passlib (used by Ansible) supports Sha algorithms with salt and round options.

tchiotludo:
I was not clear I think :smile: here is what I think. 
It will be more explicit and allow to have as many Hash as we need in a future  : 
 
```java
@Data
public class BasicAuth {
    String username;
    String password;
    PasswordHash passwordHash;
    List<String> groups = new ArrayList<>();
    
    @SuppressWarnings("UnstableApiUsage")
    public boolean isValidPassword(String password) {
        if (this.passwordHash == PasswordHash.SHA256) {
            // See http://www.mindrot.org/projects/jBCrypt/
            return BCrypt.checkpw(password, this.password);
        } else {
            // Default Sha256 format
            return this.password.equals(
                Hashing.sha256()
                    .hashString(password, StandardCharsets.UTF_8)
                    .toString()
            );
        }
        
        return this.password.equals(
                Hashing.sha256()
                        .hashString(password, StandardCharsets.UTF_8)
                        .toString()
        );
    }
    
    public enum PasswordHash {
        SHA256,
        BCRYPT
    }
}

```

gquintana:
Yes you were clear.

Except I thought you wanted to make `PasswordHash passwordHash` a settings shared for all users. I looked at the code, but couldn't find the way to do it. So I resigned to the simplest solution I could think about.

I'll implement that way.

tchiotludo:
`BasicAuth` is a [Configuration Properties](https://docs.micronaut.io/latest/guide/index.html#configurationProperties) that will be filled automatically by micronaut, just add the properties and it's done ;)

PR #428, opened 2020-10-02
427 oidc base path

### LINKED ISSUE #427: OIDC login doesn't work behind reverse Proxy
The api call to api/auths doesn't take into considering the basePath and therefore the Login button doesn't show up
image

image

yanayita:
Fixes #427 
Added basePath to endpoints.js and set configuration in application.yml to use micronaut.server.context-path


tchiotludo:
Thanks :)

PR #431, opened 2020-10-02
disable empty topic button on compact topic

### LINKED ISSUE #399: When emptying topic - request parameters do not satisfy configured policy
image
As title states - i tried emptying a topic and was greeted with this notification - 
pulled latest dev image from hours ago -

Wondering if there is additional setting needed to clear the topic.

`akhq_1                   | 2020-09-15 00:53:06,131 WARN  pGroup-1-5 org.akhq.log.access        [Date: 2020-09-15T00:53:06.10898Z] [Duration: 22 ms] [Url: DELETE /api/docker-kafka-server/topic/table_changes_configdb_config_field/data/empty] [Status: 409] [Ip: 172.28.1.106] [Length: 216] [Port: 8080]
`

polarising-java:
Fix #399

### DIFF
--- a/client/src/containers/Topic/Topic/Topic.jsx
+++ b/client/src/containers/Topic/Topic/Topic.jsx
@@ -24,7 +24,7 @@ class Topic extends Component {
     compactMessageToDelete: '',
     roles: JSON.parse(sessionStorage.getItem('roles')),
     topicInternal: false,
-    configs: {}
+    configs: []
   };
 
   tabs = ['data','partitions','groups','configs','acls','logs'];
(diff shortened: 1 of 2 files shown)


pantaoran:
@tchiotludo I would argue that this was implemented wrongly. It blocks the button correctly for topics with `cleanup.policy` set to `compact`, but it wrongly also blocks it in case of `compact,delete`.

On this line: https://github.com/tchiotludo/akhq/blob/dev/client/src/containers/Topic/Topic/Topic.jsx#L87
The result shouldn't be compared with `=== 'delete'`, but instead with something like `.contains('delete')`

Any chance we could fix that?

tchiotludo:
@pantaoran i've fix it ;)

Issue #359, opened 2020-08-13
Add last write datetime on topicks page


bochkarevnv:
Hello!
It would be nice if the topics viewing page would contain information about the datetime of the last write in this topic. Now for this information you need to go to the topic and change the sorting to newest or move the timestamp filter back until the data appears.
I don't have the skills to develop it myself.

Presto412:
I'm interested to work on this, could you please give some starter points?

tchiotludo:
You will need to add a Method `getLastRecord(String clusterId, String topic)` on `RecordRepository` and add a new route on `TopicController`

After we need to find the  logic on `RecordRepository`: 
- There is [this response](https://stackoverflow.com/questions/31735852/kafka-how-to-get-last-modified-time-for-a-topic-i-e-last-message-added-to-any?rq=1) that is very old and don't seems to be available on last client (to confirm)
- Last option I see is to use a KafkaConsumer, get last offset (`consumer.endOffsets`) and decrement offset (-1) till you have a message (`consumeSingleRecord`) and return the last record 
- Or find a better way :smile: 

Last, add a async call on the controller in order to add a col on `TopicList.jsx`

Presto412:
Thanks @tchiotludo , I'll check this out.

Presto412:
@tchiotludo I'm fairly new to kafka, but below is what I could come up with. Is this the right approach?
I get all the partitions of the topic, and then poll the messages, and decrement them based on timestamp. Please have a look

```java
           public Optional<Record> getLastRecord(String clusterId, String topicName, Options options) throws ExecutionException, InterruptedException {

        Topic topic = topicRepository.findByName(clusterId, topicName);
        List<TopicPartition> topicPartitions = topic.getPartitions().stream().map(partition -> new TopicPartition(partition.getTopic(), partition.getId())).collect(Collectors.toList());
        KafkaConsumer<byte[], byte[]> consumer = kafkaModule.getConsumer(clusterId);
        Map<TopicPartition, Long> endOffSets = consumer.endOffsets(topicPartitions);
        ConsumerRecord<byte[], byte[]> lastRecord = null;
        while (!endOffSets.isEmpty()) {
            consumer.assign(endOffSets.keySet());
            endOffSets.forEach(consumer::seek);
            ConsumerRecords<byte[], byte[]> consumerRecords = this.poll(consumer);
            if (!consumerRecords.isEmpty()) {
                while (consumerRecords.iterator().hasNext()) {
[...]

PR #439, opened 2020-10-05
Added last write to topic

Presto412:
Relevant to issue #359


src/main/java/org/akhq/repositories/RecordRepository.java
```diff
@@ -377,21 +417,29 @@ private int pollSizePerPartition(Topic topic, Options options) {
...
+                avroWireFormatConverter.convertValueToWireFormat(record, this.kafkaModule.getRegistryClient(clusterId))
+        );
+    }
+
     private Record newRecord(ConsumerRecord<byte[], byte[]> record, BaseOptions options) {
         return new Record(
-            record,
-            this.schemaRegistryRepository.getKafkaAvroDeserializer(options.clusterId),
-            avroWireFormatConverter.convertValueToWireFormat(record, this.kafkaModule.getRegistryClient(options.clusterId))
+                record,
```
tchiotludo, line 430: Noise, please remove this indent

src/main/java/org/akhq/repositories/RecordRepository.java
```diff
@@ -377,21 +417,29 @@ private int pollSizePerPartition(Topic topic, Options options) {
...
     }
 
     private RecordMetadata produce(
-        String clusterId,
-        String topic, byte[] value,
-        Map<String, String> headers,
-        byte[] key,
-        Optional<Integer> partition,
-        Optional<Long> timestamp
+            String clusterId,
```
tchiotludo, line 437: Noise, please remove this indent

client/src/containers/Topic/TopicList/TopicList.jsx
```diff
@@ -275,6 +275,22 @@ class TopicList extends Root {
               colName: 'Size',
               type: 'text'
             },
+            {
+              id: 'lastWrite',
+              accessor: 'name',
+              colName: 'Last Record Write',
+              type: 'text',
+              cell: (obj, col) => {
+                return this.getApi(uriTopicLastRecord(selectedCluster, obj[col.accessor]))
```
tchiotludo, line 284: Don't do apiCall during rendering of the view, you need to use React State, like for Consumer Group : 

https://github.com/tchiotludo/akhq/blob/9dd39982eb79f4acf2cbdebcbd90a1a8bb2871f9/client/src/containers/Topic/TopicList/TopicList.jsx#L142-L165 

We need to gather all api call in state (to have a full vision of all api call) and it will display a loading mask (done with default undefined value)

tchiotludo (changes requested):

src/main/java/org/akhq/repositories/RecordRepository.java
Presto412: ack

src/main/java/org/akhq/repositories/RecordRepository.java
Presto412: ack

client/src/containers/Topic/TopicList/TopicList.jsx
Presto412: ack

Presto412:
Hey @tchiotludo could you please check the updates? Thanks.

tchiotludo:
Hey :) 
I've a quick look and seems to be ok. 
I just need to test with real cluster, I'll try to it this week end :+1:

Presto412:
Sure, thanks!

Regards,
Priyansh Jain

tchiotludo:
Sorry, this didn't work : 
```
Internal Server Error: null
java.lang.NullPointerException
	at org.apache.kafka.clients.consumer.KafkaConsumer.seek(KafkaConsumer.java:1618)
	at java.base/java.util.HashMap.forEach(HashMap.java:1336)
	at org.akhq.repositories.RecordRepository.getLastRecord(RecordRepository.java:88)
	at org.akhq.controllers.TopicController.lastRecord(TopicController.java:189)
	at org.akhq.controllers.$TopicControllerDefinition$$exec6.invokeInternal(Unknown Source)
...
```
for all topics. 

image

To be honest, I don't know the `commited` method on consumer but it always return null on my case for every topics.
Does this method is not intended to be used with a real consumer group ?

Presto412:
Hmm, this is interesting. I don't see anything specific in the Javadoc relating to this. I'll try another approach soon.

Presto412:
Hey @tchiotludo I tried something else
I can't get akhq to work locally, so could you please test the same with your local cluster
used [this](https://stackoverflow.com/questions/57722688/is-there-a-way-to-get-the-last-message-from-kafka-topic) as reference for the implementation
Thanks!

tchiotludo:
Just try and this don't work also.
you call `consumer.endOffsets` but with no `assign` that just lead to empty result. 

You can do a unit test (that is needed for this PR) if you can't have a full akhq env working. 
The ui seems to be good, for we need to have a unit test on back end I think

Presto412:
Ok @tchiotludo , will try with unit tests

Presto412:
@tchiotludo Pushed a change .. I'm able to see a successful test when a complete gradle build is done, but I'm not able to run unit tests individually. Kindly have a look

tchiotludo:
Will look

tchiotludo:
I think for need to enable annotations processors on your ide

Presto412:
@tchiotludo It's an issue with not being able to find a kafka broker node

src/main/java/org/akhq/repositories/RecordRepository.java
```diff
@@ -61,6 +72,32 @@
     @Value("${akhq.topic-data.poll-timeout:1000}")
     protected int pollTimeout;
 
+
+    public Optional<Record> getLastRecord(String clusterId, String topicName) throws ExecutionException, InterruptedException {
```
tchiotludo, line 76:
```java
    public Optional<Record> getLastRecord(String clusterId, String topicName) throws ExecutionException, InterruptedException {
        KafkaConsumer<byte[], byte[]> consumer = kafkaModule.getConsumer(clusterId);
        Topic topic = topicRepository.findByName(clusterId, topicName);

        List<TopicPartition> topicPartitions = topic
                .getPartitions()
                .stream()
                .map(partition -> new TopicPartition(partition.getTopic(), partition.getId()))

        consumer.assign(topicPartitions);
        AtomicReference<ConsumerRecord<byte[], byte[]>> latestRecord = new AtomicReference<>();

        consumer
            .endOffsets(consumer.assignment())
            .forEach((topicPartition, offset) -> {
                consumer.seek(topicPartition, Math.max(0, offset - 3));
            });

        AtomicLong maxTimestamp = new AtomicLong();
        this.poll(consumer).forEach(record -> {
            if (record.timestamp() > maxTimestamp.get()) {
                maxTimestamp.set(record.timestamp());
                latestRecord.set(record);
            }
        });

        consumer.close();
        return Optional.ofNullable(latestRecord.get()).map(record -> newRecord(record, clusterId));
    }
```
I need to change to this to make it work with topic having transactions.
Also I poll only one time except one time per topic for performance.

src/main/java/org/akhq/controllers/TopicController.java
```diff
@@ -192,6 +192,12 @@ public Topic home(String cluster, String topicName) throws ExecutionException, I
         return this.topicRepository.findByName(cluster, topicName);
     }
 
+    @Get("api/{cluster}/topic/{topicName}/last-record")
+    @Operation(tags = {"topic"}, summary = "Retrieve the last record of a topic")
+    public Record lastRecord(String cluster, String topicName) throws ExecutionException, InterruptedException {
+        return this.recordRepository.getLastRecord(cluster, topicName).orElse(null);
```
tchiotludo, line 198:
```suggestion
        return this.recordRepository.getLastRecord(cluster, topicName).orElse(new Record());
```
To avoid 404 on topic with no data

client/src/containers/Topic/TopicList/TopicList.jsx
```diff
@@ -185,11 +186,16 @@ class TopicList extends Root {
...
-          setState()
-        })
-    });
+          .then(value => {
+            tableTopics[topic.name].groupComponent = value.data
+            setState()
+          });
+      this.getApi(uriTopicLastRecord(selectedCluster, topic.name))
+          .then(value => {
+            tableTopics[topic.name].lastWrite = value.data.timestamp
```
tchiotludo, line 195:
```suggestion
            tableTopics[topic.name].lastWrite = value.data.timestamp || ''
```
To remove loading box on empty topic

client/src/containers/Topic/TopicList/TopicList.jsx
```diff
@@ -317,6 +323,12 @@ class TopicList extends Root {
               colName: 'Size',
               type: 'text'
             },
+            {
+              id: 'lastWrite',
+              accessor: 'lastWrite',
+              colName: 'Last Record Write',
```
tchiotludo, line 329:
```suggestion
              colName: 'Last Record',
```

tchiotludo (changes requested):
Almost working. 
Final Feature to add : ability to display record on popup ! 
We all the data, it will be nice to allow display adding a link the date and opening a single row of data view : 
image

src/main/java/org/akhq/repositories/RecordRepository.java
Presto412: Hey @tchiotludo , Why should it be offset - 3?

tchiotludo: -2 must be sufficient I think. 
Kafka transaction commit an offset that the consumer can't read, if you try with -1 you will never get the last message

Presto412: how to determine if topic is transactional? We can add an if check for that - if transactional, give offset -2 else 1

tchiotludo: You can't, it's only known by the producer and a topic can contains transactions or not, no way to know it

Presto412: then we can check both offsets and return a value if found

Presto412: Could you give example for opening the linked record in a view? Will try to incorporate that change as well

Presto412:
@tchiotludo Added the offset difference change, kindly check

tchiotludo:
why have changed my version with only -2 with this version  ? 
```java
        Optional<ConsumerRecord<byte[], byte[]>> latestRecord =
            getLatestRecordFromOffsetDifference(consumer, 1)
                .or(() -> getLatestRecordFromOffsetDifference(consumer, 2));
```
You are polling 2 time the cluster, and it will be slower than my simple version (poll 2 message and only keep the last one)

Presto412:
Oh my bad, I assumed that we need to poll each offset, and doing offset -2
won't consume the offset - 1 message. Will change.

Regards,
Priyansh Jain

tchiotludo:
I've merged your PR on b1558278113bb1c5549cd92cf5fd7bdafcf6989b and improve it on 4f6f23b2b652f966cf7611348ead29f44ede6379.

Thanks for the contributions

PR #442, opened 2020-10-07
Run Container as a non-root user by default and update the Helm chart

chgl:
Run container as user 10001 by default. Added options for setting the (pod) securityContext. Added support for ingress v1, if available.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,5 +3,6 @@ FROM openjdk:11-jre-slim
 WORKDIR /app
 COPY docker /
 ENV MICRONAUT_CONFIG_FILES=/app/application.yml
+USER 10001
 ENTRYPOINT ["docker-entrypoint.sh"]
 CMD ["./akhq"]
--- a/helm/akhq/Chart.yaml
+++ b/helm/akhq/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.0"
 description: Kafka GUI for Apache Kafka to manage topics, topics data, consumers group, schema registry, connect and more...
 name: akhq
-version: 0.1.2
+version: 0.1.3
 keywords:
   - kafka
   - confluent
--- a/helm/akhq/templates/deployment.yaml
+++ b/helm/akhq/templates/deployment.yaml
@@ -37,6 +37,8 @@ spec:
       imagePullSecrets:
         {{ toYaml .Values.imagePullSecrets | indent 8 }}
       {{- end }}
+      securityContext:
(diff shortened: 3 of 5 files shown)


Dockerfile
```diff
@@ -3,5 +3,6 @@ FROM openjdk:11-jre-slim
 WORKDIR /app
 COPY docker /
 ENV MICRONAUT_CONFIG_FILES=/app/application.yml
+USER 10001
```
tchiotludo, line 6: I think it's better to create a real user, no ?

helm/akhq/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.0"
 description: Kafka GUI for Apache Kafka to manage topics, topics data, consumers group, schema registry, connect and more...
 name: akhq
-version: 0.1.2
+version: 0.1.3
```
tchiotludo, line 5: Helm chart are generated from master branch, this version already exist

tchiotludo (changes requested):

helm/akhq/Chart.yaml
chgl: Ah, apologies. Should I create a new PR with the helm changes to master and the Dockerfile changes to dev?

Dockerfile
chgl: It generally doesn't matter. Eg OpenShift just randomizes the user id a container is run as. The advantage of using a numeric ID (independent of whether the user actually exists, i.e. has a passwd entry), is that Kubernetes can tell if a container is run as a non-root user because the number is != 0 (see also https://stackoverflow.com/questions/49720308/kubernetes-podsecuritypolicy-set-to-runasnonroot-container-has-runasnonroot-and)

helm/akhq/Chart.yaml
tchiotludo: Yes it's better :) I need to provide a doc about this ;)

Dockerfile
chgl: Yep! I've built a custom image `FROM akhq` here: https://hub.docker.com/layers/chgl/akhq/0.15.0-rootless/images/sha256-daf750aa2e94e1c3a75b91faf5df08bdda4d6ef8dcf70cd8df7e02cc8b2acfd5?context=explore and it works fine. I'm currently running the official container image within a Kubernetes cluster but overriding the securityContext (after realizing the chart already provides this feature) to a non-root user. No problems with either approach, as the default permissions for config and jar allow read-access to any user.

PR #445, opened 2020-10-10
Publish distribution Tar and Zip

gquintana:
I am a complete Gradle and Github Action newbie. I don't know 
* how to get Gradle project version from Github workflow?
* how to test this change?
* Tar, Zip or both?


tchiotludo:
Seems to be ok ;) 
Will wait for next release to validate it :)

PR #458, opened 2020-10-16
WIP: copy topic data

polarising-java:


tchiotludo:
Hi ! 
To be honest, I'm a little bit confused about the way you implement it, I think you have a switch some input on the ui.
When I click on copy, I've this screen : 
image
So I expect to have offset from the topic I select in first, not after like that. 
image

I expect: 
-  choose the topic I want to copy from
- choose the offset 
- Choose the destination cluster 
- Choose the destination topic

I am wrong ? 

Also I really think this filter is quite mandatory : 
image
maybe with another one, x message per partition, most of people will want to have a subset of a topic to another cluster I think.

tchiotludo:
almost perfect, just add minor change 

Thanks !!

Issue #460, opened 2020-10-20
Sort topic data OLDEST / NEWEST


PLarboulette:
Hello, 

I think the 
`  topic-data:
    sort: NEWEST
`
is broken. 

Even if the parameter is set to NEWEST, topics are sorted by the default value, OLDEST.

tchiotludo:
Seems that you don't respect yaml indent.
Can you paste the whole file please ?

PLarboulette:
```
akhq:
  server:
    access-log:
      enabled: true
      name: org.akhq.log.access
      format: "[Date: {}] [Duration: {} ms] [Url: {} {} {}] [Status: {}] [Ip: {}]
        [Length: {}] [Port: {}]"
  clients-defaults:
    consumer:
      properties:
        isolation.level: read_committed
  connections:
    kafka-cluster:
      properties:
        bootstrap.servers: "${KAFKA_BOOTSTRAP_SERVERS}"
      schema-registry:
        url: "${SCHEMA_REGISTRY_URL}"
      connect:
        - name: "${KAFKA_CONNECT_NAME}"
          url: "${KAFKA_CONNECT_URL}"
  pagination:
    page-size: 25
    threads: 16
  topic-data:
    sort: NEWEST
    size: 50
    poll-timeout: 1000
```

topic and pagination are in the same levels as connections or server. When I look the example file, it seems to be the same thing.

tchiotludo:
Ok got it :) 
This option is not transmitted to client so It can't work :) 

Good catch :+1:

PR #464, opened 2020-10-20
[RFC] Consumer groups on the topic listing with the new client

ankon:
In akhq up to 0.15.0 (with the old UI) it was possible to prevent akhq from querying the consumer groups for all topics, as that could easily be very very expensive, and in those cases is usually not relevant. 

With the old UI gone the `akhq.topic.skip-consumer-group` setting is no longer relevant, as the frontend will simply query the consumer groups for each topic directly. Therefore #463 removes that setting.

But that leaves a hole: Letting akhq query the consumer groups in my environment will just ensure that the Kafka brokers get unhappy and consume a lot of CPU cycles, the network gets unhappy by being filling up with garbage, and I get unhappy because eventually akhq's UI just shows errors about things I knew would fail anyways :) 

This PR is definitely not for merging, but I think it's good to discuss how akhq could be made to work in such environments with thousands of consumer groups.

What could work for us could for example be to provide some form of settings in the `application.yml` that tell the frontend to not query the consumer groups. But then comes the question of how to get the settings from the backend to the frontend -- maybe a new `/api/client-settings` call, or `/client/settings` that gets handled by the backend? 

WDYT?

### DIFF
--- a/client/src/containers/Topic/TopicList/TopicList.jsx
+++ b/client/src/containers/Topic/TopicList/TopicList.jsx
@@ -5,13 +5,12 @@ import Header from '../../Header';
 import SearchBar from '../../../components/SearchBar';
 import Pagination from '../../../components/Pagination';
 import ConfirmModal from '../../../components/Modal/ConfirmModal';
-import { uriDeleteTopics, uriTopics, uriTopicsGroups } from '../../../utils/endpoints';
+import { uriDeleteTopics, uriTopics } from '../../../utils/endpoints';
 import constants from '../../../utils/constants';
 import { calculateTopicOffsetLag, showBytes } from '../../../utils/converters';
 import './styles.scss';
 import { toast } from 'react-toastify';
 import 'react-toastify/dist/ReactToastify.css';
-import {Collapse} from 'react-bootstrap';
 import Root from '../../../components/Root';
 
 class TopicList extends Root {
@@ -39,7 +38,6 @@ class TopicList extends Root {
     roles: JSON.parse(sessionStorage.getItem('roles')),
     loading: true,
     cancel: undefined,
-    collapseConsumerGroups: {}
   };
 
   componentDidMount() {
@@ -161,13 +159,6 @@ class TopicList extends Root {
 
   handleTopics(topics) {
     let tableTopics = {};
-    const collapseConsumerGroups = {};
-
-    const { selectedCluster } = this.state;
-
-    const setState = () =>  {
-      this.setState({ topics: Object.values(tableTopics) });
-    }
 
     topics.forEach(topic => {
       tableTopics[topic.name] = {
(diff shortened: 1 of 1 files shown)


twobeeb:
Hi, I agree with @ankon this is a feature that have a big impact on performance. 

As I moved to 0.16.0, I decided to deploy along with AKHQ a new Ingress rule that redirects _api/{cluster}/topic/{topicName}/groups_ to a different service than AKHQ (mmock) that replies an empty json list "[]" whenever called. 

I too would rather have a AKHQ solution to this problem.
Maybe load consumer groups on demand, by topic ?

tchiotludo:
Hey, 
You don't seems to be the only to request it : https://github.com/tchiotludo/akhq/issues/466
Also there is a need to have some config send from backend to frontend : https://github.com/tchiotludo/akhq/issues/460

So clearly, we need to add some endpoint on backend to forward configuration.
I will suggest to have some `akhq.ui.*` special option to forward options to client with an endpoint `/api/ui-options` that will transmit this options to client. 

The last thing we need to think, is what feature-flag we need. 

In your case, the consumer group page will also have a big impact on performance ? I'm wrong ? how do you deal with this ?

tchiotludo:
in a mean time, in commit a change on `dev` branch https://github.com/tchiotludo/akhq/commit/dde81148e47cafba9f398562def55656ee87ee2a that can fix performance issue. 
I don't know if it can help for your use case ?

ankon:
> I don't know if it can help for your use case ?

We have probably a couple of hundred topics, and most of these have on the order of a couple of hundred consumer groups (99.9% of which are dead). For us this isn't a problem of "improving" as much as it really is "we don't care about the consumer groups at all" :)

ankon:
> As I moved to 0.16.0, I decided to deploy along with AKHQ a new Ingress rule that redirects _api/{cluster}/topic/{topicName}/groups_ to a different service than AKHQ (mmock) that replies an empty json list "[]" whenever called.

Neat idea! I had the development environment etc for akhq already around, so building a private version that just dropped the calls completely was easier. But I'll keep the mmock idea in the back of my mind :)

> I will suggest to have some akhq.ui.* special option to forward options to client with an endpoint /api/ui-options that will transmit this options to client.

Sounds reasonable to me!

tchiotludo:
@ankon I understand your point of view, and since consumer group on topics list don't have any interest, the better will be to some feature flag for the ui.
User want a lot of feature on ui (example: https://github.com/tchiotludo/akhq/pull/439) that can use a lot of resource of some kafka cluster, so a feature flag per cluster  `/api/{cluster}/ui-options` can be fine !
You do you think about it ?

tchiotludo:
@ankon @twobeeb this feature is merged based on #511, tell me if it works for you

ankon:
This definitely looks very promising, thanks a lot! Will try to give it a spin over the coming days, but I suspect already that this will be exactly what we need :)

ankon:
@tchiotludo and @polarising-java: Thanks a lot! Indeed, for me this seems to work just fine.

Issue #254, opened 2020-03-30
Feature request: Topic configurations by cluster


samcarpentier:
### Description

Given that AKHQ allows to manage multiple clusters, it would be great to be able to specify topics configurations by cluster.

The most striking example of this is the following: I have 2 clusters registered in AKHQ. One has 12 nodes, the other has 6 nodes. I want to be able to specify a default replication factor that's different for each cluster.

See the proposed configuration example below.

### Current configuration

```yaml
# ...
  connections:
    kafka-cluster1:  # 12 nodes
      properties:
        bootstrap.servers: kafka-cluster1.example.com:9092
    kafka-cluster2:  # 6 nodes
      properties:
        bootstrap.servers: kafka-cluster2.example.com:9092

  topic:
    replication: 4
    partitions: 3
# ...
```

### Proposed configuration

```yaml
# ...
  connections:
    kafka-cluster1:  # 12 nodes
      properties:
        bootstrap.servers: kafka-cluster1.example.com:9092
    kafka-cluster2:  # 6 nodes
      properties:
        bootstrap.servers: kafka-cluster2.example.com:9092

  topic:
    kafka-cluster1:
      replication: 8
      partition: 3
    kafka-cluster2:
      replication: 4
      partition: 3
# ...
```

samcarpentier:
@tchiotludo Let me know what you think about this and what would be the effort to implement this.

tchiotludo:
not really complicated, just need to think the proper way to deal with the configuration files. 
maybe add a `defaults:` to be clear that's the default one, and allow override on the `connections` part.
Not really sure right now :thinking:

samcarpentier:
Oh, that's a good point. Didn't think of that. So it would look something like this:

```yaml
# ...
  connections:
    kafka-cluster1:  # 12 nodes
      properties:
        bootstrap.servers: kafka-cluster1.example.com:9092
    kafka-cluster2:  # 6 nodes
      properties:
        bootstrap.servers: kafka-cluster2.example.com:9092

  topic:
    default:
      retention: 172800000
      partition: 3
      default-view: HIDE_INTERNAL
      internal-regexps:
        - "^_.*$"
        - "^.*_schemas$"
        - "^.*connect-config$"
        - "^.*connect-offsets$1"
        - "^.*connect-status$"
      stream-regexps:
        - "^.*-changelog$"
        - "^.*-repartition$"
        - "^.*-rekey$"
      skip-consumer-groups: false
[...]

PR #477, opened 2020-10-26
New attribute to filter Connect Configs by regexp

twobeeb:
I followed the same logic you implemented for topics.
It has the sames limitations as your implementation : it only makes sense when the user have Read-Only acces, because the filter is only verified on list and search.
It has the same vulnerability as your implementation : if the user have roles read/create/update/delete, he can create/update/delete connect configs that doesn't follow the regex.

I consider this OK because for our use cases, we use AKHQ as a Read Only support/monitoring tool.
I may be worth mentioning this behavior in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,6 +336,7 @@ Define groups with specific roles for your users
   * `- name: group-name` Group identifier
     * `roles`: Roles list for the group
     * `attributes.topics-filter-regexp`: Regexp to filter topics available for current group
+    * `attributes.connects-filter-regexp`: Regexp to filter Connect tasks available for current group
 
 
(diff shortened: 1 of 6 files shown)


tchiotludo:
Seems to be fine ! 
Can you just add a unit test to validate it please ? 
I will not use this feature myself, I would prefer to have a unit test to keep it in my mind. 
Also we need to move this when doing #254, will have a unit to not forgot

tchiotludo (changes requested):
see previous comment

tchiotludo:
Thanks a lot :)

Issue #470, opened 2020-10-21
Default JWT secret


twobeeb:
The default behavior when micronaut.security.token.jwt.signatures.secret.generator.secret is not set is to generate a JWT without signature, that lets anyone change its claim effortless.

I believe this is a security flaw and AKHQ should not even start if there is no secret, or at least there should be a default secret, along with a warning log at run time when the secret have not been changed.

tchiotludo:
I think Micronaut will crashed on this case,  don't know if we need to report it to micronaut or handle it on akhq

twobeeb:
For reference, what another project did (not using micronaut though)
https://github.com/gravitee-io/gravitee-management-rest-api/blob/master/gravitee-rest-api-management/gravitee-rest-api-management-security/src/main/java/io/gravitee/rest/api/management/security/config/BasicSecurityConfigurerAdapter.java#L179-L197

tchiotludo:
Got idea, I'll start like that :) 
Thanks

PR #478, opened 2020-10-26
Add default invalid JWT secret

twobeeb:
Fix for #470 

When launching AKHQ with a default and invalid secret (changeme), I get an authentication Error and no JWT generated.
Reading the logs, I understand it is mandatory to change the JWT.

````
2020-10-26 09:08:51,142 WARN  1-thread-6 .s.t.j.v.JwtTokenValidator JOSEException while generating token The secret length must be at least 256 bits  
2020-10-26 09:08:51,143 WARN  1-thread-6 org.akhq.log.access        [Date: 2020-10-26T09:08:47.24114Z] [Duration: 3902 ms] [Url: POST /login] [Status: 500] [Ip: /198.18.143.226
] [User: Anonymous]
````

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -33,6 +33,10 @@ micronaut:
         cookie:
           enabled: true
           cookie-same-site: strict
+        signatures:
+          secret:
+            generator:
+              secret: changeme
     redirect:
       login-success: "${micronaut.server.context-path:}/ui"
       forbidden:


tchiotludo:
To be honest, I would prefer a proper warm on console like gravitee as you suggest on #470 .

Maybe with a [PostConstruct](https://docs.micronaut.io/latest/guide/index.html#lifecycle) annotation ?

twobeeb:
I understand. 
While I believe it's more vulnerable to attacks, a default secret and a warning during startup is a good step up (signed token at least)
I close this PR.

PR #487, opened 2020-10-31
Run image as a non-root user by default

chgl:
Follow up from #442, using the correct branches this time 😅.

The changes were tested using a custom image build (see https://hub.docker.com/layers/chgl/akhq/0.15.0-rootless/images/sha256-daf750aa2e94e1c3a75b91faf5df08bdda4d6ef8dcf70cd8df7e02cc8b2acfd5?context=explore) and akhq is currently running as a non-root user set via securityContext in a production cluster.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,5 +3,6 @@ FROM openjdk:11-jre-slim
 WORKDIR /app
 COPY docker /
 ENV MICRONAUT_CONFIG_FILES=/app/application.yml
+USER 10001
 ENTRYPOINT ["docker-entrypoint.sh"]
 CMD ["./akhq"]


tchiotludo:
Sorry for the noise about branch but Docker image is better on `dev` & helm chart on `master`. 
I've added the non root on dev to avoid to publish a new `latest` docker image

thomas-chauvet:
Hi,

Apologies in advance if this is not the right place to comment (I am not  a real open-source contributor).

To avoid a user with no name it would be probably better to create a user first.

In docker, with a user with no name you will end with `I have no name!@5af14cf2a980:/app$`.

One solution could be:

```Dockerfile
FROM openjdk:11-jre-slim

WORKDIR /app
COPY docker /
ENV MICRONAUT_CONFIG_FILES=/app/application.yml

# Create user
RUN useradd -ms /bin/bash akhq

# Chown to write configuration
RUN chown -R akhq /app

USER akhq

ENTRYPOINT ["docker-entrypoint.sh"]
CMD ["./akhq"]
```

Happy to make a PR if needed :)

tchiotludo:
I agree also :) PR are welcome :+1:

PR #492, opened 2020-11-03
Add TVG to Whos using AKHQ

senzzzi:
In TVG we are using AKHQ in every environment. For us, it's a must have on the Development and QA phases and also to troubleshoot any potential issue in production. It already has LDAP authentication which we use. We don't use the multiple cluster feature, we just deployed one AKHQ per environment. Feel free to also add our logo to the website. 

Thanks a lot for your and all the contributors efforts on this. Will, for sure, recommend and continue to use it in the future.


tchiotludo:
Thanks for your report !! 
It's still cool to have some true report like that for maintainer to give some sense working for oss project. 

Really appreciate ! :pray:

PR #494, opened 2020-11-04
feat(dev): Run docker image as non-root with 'akhq' user

thomas-chauvet:
Following pull request #487,to avoid a user with no name we can create a user 'akhq' with proper rights on `/app` directory to write configuration.

The previous configuration use user `10001` which lead to an unnamed user (e.g. `I have no name!@5af14cf2a980:/app$`).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,11 @@ FROM openjdk:11-jre-slim
 WORKDIR /app
 COPY docker /
 ENV MICRONAUT_CONFIG_FILES=/app/application.yml
-USER 10001
+# Create user
+RUN useradd -ms /bin/bash akhq
+# Chown to write configuration
+RUN chown -R akhq /app
+# Use the 'akhq' user
+USER akhq
 ENTRYPOINT ["docker-entrypoint.sh"]
 CMD ["./akhq"]


chgl:
I'm perfectly fine with this change but just to give my motivation behind #487 :

Does the akhq user really need to write config? At least from my usage it seems like it will only ever read them. Having the filesystem as read-only is also solid security precaution. You can of course still give permissions to an unnamed user: `chown -R 10001:10001 /app` works without a passwd entry.

When running inside Kubernetes a numeric value allows the cluster to validate if the container is running as a non-root user, which is impossible to do if it is named. Platforms like Openshift additionally randomize the UID.

> The previous configuration use user 10001 which lead to an unnamed user (e.g. I have no name!@5af14cf2a980:/app$).

One could login as root `docker exec -u 0 -it` if necessary. But regularly logging in to containers is generally advised against. From a paranoid security perspective: base images like distroless don't include a shell at all.

thomas-chauvet:
> Does the akhq user really need to write config?

Yes, in my case I needed when I give the AKHQ configuration via the `AKHQ_CONFIGURATION` env variable.

> When running inside Kubernetes a numeric value allows the cluster to validate if the container is running as a non-root user, which is impossible to do if it is named. Platforms like Openshift additionally randomize the UID.

Didn't know this, good to know when I'll move to OpenShift :)

We can close this if needed, I'll use my own Dockerfile on top of the one provided.

chgl:
> AKHQ_CONFIGURATION

Hadn't considered that! Then let's keep this pr open. And save you the headache of maintaining your own docke file! I'm cool with anything but root by default 😄

tchiotludo:
I've just the bug of `AKHQ_CONFIGURATION` on my local env, this one is needed :+1: 

Thanks :)

PR #499, opened 2020-11-07
Partition auto assign when producing to topic

xpr3sso:
Reintroducing the option of letting Kafka library decide where to put a message based on it's key. 

This was possible with the old front end, when leaving partition selector blank on producing a message to a topic. To have this feature back again, and make it's usage more intuitive I though of adding the option "auto assign":

image

What do you say?


tchiotludo:
make sense :) 
Mostly a forgotten feature during react rework. 

Thanks a lot :+1:

PR #511, opened 2020-11-13
WIP: add ui options endpoint

polarising-java:


polarising-java:
add the structure below each connection:

`
options:
        topic:
          default-view: ALL
          skip-consumer-groups: false
        topic-data:
          sort: NEWEST
`

like :
`
  connections:
    my-cluster-plain-text:
     .....
      options:
        topic:
          default-view: ALL
          skip-consumer-groups: false
        topic-data:
          sort: NEWEST
`

If present overrides default values.

polarising-java:
In the client side added a settings page to override values in the localstorage

polarising-java:
Added skip last record date setting

polarising-java:
Also fix the change cluster while topic list is loading, and checked the time to get the last record date and is around 5sec wihle for consumer groups is around 30sec ( in the aiven cluster)

PR #512, opened 2020-11-14
feat(helm): add option to specify serviceAccountName

tarvip:
At the moment it is possible to specify init containers, but if you need serviceAccount for init container, then there is no way to specify it.


tchiotludo (changes requested):
Can you just add the properties on `values.yaml` to let people that this one exist please ?

tarvip:
> Can you just add the properties on values.yaml to let people that this one exist please ?

Updated.

tchiotludo:
Thanks 👍

Issue #524, opened 2020-11-24
Login not authorized when a user belong to two groups


apellegr06:
Hi,

I have the ldap authentification configured and I have an issue since I update from 0.12.0 to 0.16.0

My user belong to two groups defined into the yml and the connection issue is related to that.
If I remove one of the groups than I can connect, otherwise I have an error "Wrong Username or Password!"
If I rollback to 0.12.0 I don't have this issue.

Regards
Alain

tchiotludo:
Can you share the configuration files please to have a direct reproduction please ?

apellegr06:
Here is my config :

micronaut:
  server:
    port: 8081
  security:
    enabled: true
    ldap:
      default:
        enabled: true
        context:
          server: 'ldaps://xxxxxxxxxx:636'
          managerDn: 'uid=xxxxxxxxxx,ou=xxxxxxxxxxxxxx,dc=xxxxxxxxx,dc=xx'
          managerPassword: 'xxxxxxx'
        search:
          base: 'dc=xxxxxxxxx,dc=xx'
        groups:
          enabled: true
          base: 'ou=xxxxx,dc=xxxxxxxxx,dc=xx'

akhq:
  server:
    base-path: "" # if behind a reverse proxy, path to kafkahq without trailing slash (optional). Example: kafkahq is
                  # behind a reverse proxy with url http://my-server/kafkahq, set base-path: "/kafkahq".
                  # Not needed if you're behind a reverse proxy with subdomain http://kafkahq.my-server/
    access-log: # Access log configuration (optional)
      enabled: true # true by default
      name: org.akhq.log.access # Logger name
      format: "[Date: {}] [Duration: {} ms] [Url: {} {}] [Status: {}] [Ip: {}] [User: {}]" # Logger format

  # default kafka properties for each clients, available for admin / producer / consumer (optional)
  clients-defaults:
    consumer:
      properties:
        isolation.level: read_committed
        default.api.timeout.ms: 60000

  # list of kafka cluster available for kafkahq
  connections:
# ---- [ ADD CLUSTER CONNECTION AFTER THIS LINE ] ----
    XXXXXXXXXXXXX:
      properties:
        bootstrap.servers: "xxxxxxxxxxxxxxxxxxxxxx"
        security.protocol: SASL_PLAINTEXT
[...]

PR #526, opened 2020-11-24
fix(auth): Rolling back group configuration to a map format

ronrother:
Hi @tchiotludo 

This is a proposed fix for an issue with 0.16.0 that happens when you configure security groups (akhq.security.groups). I believe it's related to [this issue](https://github.com/tchiotludo/akhq/issues/524).

The root cause is, since the groups are now configured as a list (and not a map) the configuration that takes precedence (according to the Micronaut rules for precedence) will override all other configurations for that key. In other words, if you configure any group under `akqh.security.groups`, the default `admin`, `reader` and `no-roles` groups (defined in the resources/application.yml) will be overriden.

**A mitigation** for the users waiting for a fix is to add the basic groups to their own configuration (basically copy them from https://github.com/tchiotludo/akhq/blob/dev/src/main/resources/application.yml).

**The ideal solution** in my opinion would be to revert to using a map instead of a list for configuring groups, but I understand that might be troublesome for you.

**The solution in this PR** is a middle-ground. It renames the default groups configured in `resources/application.yml` to `basic-groups` and merges them with the ones configured by the user. There is a limitation though, only one environment is supported (https://docs.micronaut.io/latest/guide/index.html#environments). This should cover 90% if not all use cases, though, and is already a limitation in the current 0.16.0.

Hope this is helpful and apologies for not introducing myself first.

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -132,40 +132,40 @@ akhq:
 
   security:
     default-group: admin
-    basic-groups:
-    - name: admin
-      roles:
-      - topic/read
-      - topic/insert
-      - topic/delete
-      - topic/config/update
-      - node/read
-      - node/config/update
-      - topic/data/read
-      - topic/data/insert
-      - topic/data/delete
-      - group/read
-      - group/delete
-      - group/offsets/update
-      - registry/read
-      - registry/insert
-      - registry/update
-      - registry/delete
-      - registry/version/delete
-      - acls/read
-      - connect/read
-      - connect/insert
-      - connect/update
-      - connect/delete
-      - connect/state/update
-    - name: reader
-      roles:
-      - topic/read
-      - node/read
-      - topic/data/read
-      - group/read
-      - registry/read
-      - acls/read
-      - connect/read
-    - name: no-roles
-      roles: []
+    groups:
+      admin:
+        roles:
+        - topic/read
+        - topic/insert
+        - topic/delete
+        - topic/config/update
+        - node/read
+        - node/config/update
+        - topic/data/read
+        - topic/data/insert
+        - topic/data/delete
+        - group/read
+        - group/delete
+        - group/offsets/update
+        - registry/read
+        - registry/insert
+        - registry/update
+        - registry/delete
+        - registry/version/delete
+        - acls/read
+        - connect/read
+        - connect/insert
+        - connect/update
+        - connect/delete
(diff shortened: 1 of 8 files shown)


tchiotludo:
Hi @ronrother, 
Sorry for the late reply !  and thanks for this PR ! 
 
I think there is a lot of issue about that on last release ! 
To be honest, i was not aware that micronaut will overwrite all the key from on array ! 
I was think it twice before using array in configuration. 

To be transparent, knowing that, I will think reverting to map will be the best option in a long term ! 

Do you want to move to this PR to a rollback to a map ? 

If yes, just on this map the key of the map to a special properties `key` and not the `name` [here]( 
https://github.com/tchiotludo/akhq/blob/dev/src/main/java/org/akhq/configs/Group.java).
There is a lot of issue with ldap since micronaut will change the case of the key. 

WHat do you think about that ?

ronrother:
Hi @tchiotludo 

Sure, I will into that still this week.

ronrother:
Hey @tchiotludo 

I modified this PR to make the group configuration use a map instead of a list, as you requested. Note that the users that switched to the list format will have to rewrite their configurations again once this change is picked up in a release.

Sorry for the late reply.

src/main/resources/application.yml
```diff
@@ -133,39 +133,39 @@ akhq:
...
-      - topic/read
-      - node/read
-      - topic/data/read
-      - group/read
-      - registry/read
-      - acls/read
-      - connect/read
-    - name: no-roles
-      roles: []
+      admin:
```
tchiotludo, line 136: Just a minor changes here : 
```yaml
admin: 
  name: Admin
   roles: 
```
And everywhere else. 

We need to have a name & a key here because people with LDAP have group case sensitive. 
I have a lot of issues about that in the past that lead me to go to array (that was not a good idea at the end).

tchiotludo (changes requested):
Can you just add the name & the key please ? 

I know for the configuration files changed ... I love to have issue about that :thinking: but a mistake was done and need to be corrected. 

Thanks :+1:

src/main/resources/application.yml
ronrother: I added a name property to the default groups with the same name as the key. I also changed the PostConstruct of `SecurityProperties` so that, when a name is not specified in the configuration, the key will be used as default.

PR #543, opened 2020-12-16
Support ANY field in Protobuf descriptors

xakassi:
This a fix for a bug described in [PR 529](https://github.com/tchiotludo/akhq/pull/529).

`Any` field is now supported for Protobuf descriptors. Multiple layer complex objects with `Any` fields on any layers are also supported using recursion. Please, see unit tests for examples.

The principle of operation is as follows. If a message contains a field with type `Any`, we can find out an actual field type at the runtime from the message, get descriptors for this field type and add them to TypeRegistry. To get the descriptors for this field type I pre-built a map with pairs <type name> - <descriptors> for all types provided in all descriptors.

Hi, @tchiotludo , @jorgheymans ! Please, take a look at this PR.
@jorgheymans I hope you case is fully covered now, please, test on your data.


jorgheymans:
Going to test this out hopefully today, feedback soon.

jorgheymans:
Looking at this, the proto serde is not working and i'm not seeing any error messages. This leads me to believe it's a configuration error. Can you confirm this config should be correct:

```
    deserialization:
      protobuf:
        descriptors-folder: "/opt/akhq/descriptors"
        topics-mapping:
          - topic-regex: "cc.*"
            descriptor-file: "cc-message-proto.protobin"
            value-message-type: "cc.message.CCMessage"
          - topic-regex: "han.*"
            descriptor-file: "han-cdc-proto.protobin"
            value-message-type: "han.message.HanMessage"
```

Note that the descriptor files are not self-containing as it doesn't make sense once you have `Any`. So the `/opt/akhq/descriptors` directory contains many protobin files, I am assuming you scan them all ?

So basically we have `KafkaMessage.proto`

```
message KafkaMessage {
    ..........
    Any payload
}
````
And then many payload types, each type has its own protobin file. All these protobin files are in the `descriptors-folder`.

**Comment by xakassi** [comment]
Could you, please, provide the whole config file? Maybe you have placed deserialization block in the wrong place? It should be under your Kafka cluster on the same level as its "properties". I will show you my whole config once I'm at the computer. -------С уважением, Таисия Дорошенкова-------- Исходное сообщение --------Тема: Re: [tchiotludo/akhq] Support ANY field in Protobuf descriptors (#543)От: Jorg Heymans Кому: tchiotludo/akhq Копия: Taisiia Goltseva ,Author 
Looking at this, the proto serde is not working and i'm not seeing any error messages. This leads me to believe it's a configuration error. Can you confirm this config should be correct:
    deserialization:
      protobuf:
        descriptors-folder: "/opt/akhq/descriptors"
        topics-mapping:
          - topic-regex: "cc.*"
            descriptor-file: "cc-message-proto.protobin"
            value-message-type: "cc.message.CCMessage"
          - topic-regex: "han.*"
            descriptor-file: "han-cdc-proto.protobin"
            value-message-type: "han.message.HanMessage"

Note that the descriptor files are not self-containing as it doesn't make sense once you have Any. So the /opt/akhq/descriptors directory contains many protobin files, I am assuming you scan them all ?
So basically we have KafkaMessage.proto
message KafkaMessage {
    ..........
    Any payload
}

And then many payload types, each type has its own protobin file. All these protobin files are in the descriptors-folder.

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/tchiotludo/akhq/pull/543#issuecomment-748099015",
"url": "https://github.com/tchiotludo/akhq/pull/543#issuecomment-748099015",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

**Comment by xakassi** [comment]
Hi, @jorgheymans !
This is my whole config:
```
akhq:
  connections:
    kafka:
      properties:
        bootstrap.servers: "localhost:9094"
      deserialization:
        protobuf:
          descriptors-folder: "D:\\protobuf_desc"
          topics-mapping:
            - topic-regex: "test.*"
              descriptor-file: "streaming.desc"
              key-message-type: "Row"
              value-message-type: "Envelope"
```

**Comment by jorgheymans** [comment]
Indeed the config block was placed under topics, i have corrected it and as soon as I'm able to test this again i'll let you know, may take a few days with all the holidays going on.

**Comment by tchiotludo** [comment]
@jorgheymans : please tell me what you test is going on. 
I would like to release a new AKHQ version will a full support of protobuf (v1).
Thanks for your time @xakassi & @jorgheymans !

**Comment by jorgheymans** [comment]
It won't be before next week, full holiday mode here sorry !

**Comment by tchiotludo** [comment]
@jorgheymans : have a good holidays ! 
see you next week !

**Comment by xakassi** [comment]
Hi, @jorgheymans !
Did you have any chance to test it yet?

**Comment by dhakerk** [comment]
Hi,following Jorg inputs test done and we got an exception here I share with you what we got  :
proto_exception

cordially.

**Comment by xakassi** [comment]
Hi!
Are your descriptor files ok? It's an exception from standard function `FileDescriptorSet.parseFrom(descriptorFile)`, it cannot parse your descriptor file.
Could you provide more info about your descriptor files and data? Maybe some fake data to reproduce.

**Comment by jorgheymans** [comment]
It's missing the google.protobuf.timestamp descriptor, this is part of the
well known types
https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/timestamp

I guess we should generate a separate descriptor for these internal types
as well ?

**Comment by xakassi** [comment]
I have added a unit test for an object with Timestamp field. @jorgheymans , please, see the unit test `deserializeFilm()`. Film object now looks like this:
```
syntax = "proto3";
package org.akhq.utils;

import "google/protobuf/timestamp.proto";

option java_package = "org.akhq.utils";
option java_outer_classname = "FilmProto";

message Film {
  string name = 1;
  string producer = 2;
  int32 release_year = 3;
  int32 duration = 4;
  repeated string starring = 5;
  google.protobuf.Timestamp timestamp = 6;
}
```

And it deserialized successfully. 
Do your objects looks like this also? Or I have specified Timestamp in a different way?

**Comment by jorgheymans** [comment]
The error surfaces if you have a `timestamp` field in a type that is defined as `Any`.

**Comment by xakassi** [comment]
Yes, indeed. But in my case (when I test) I get other error in a different function (not here `FileDescriptorSet.parseFrom(descriptorFile)`).

But anyway my current approach of collecting necessary descriptors is incomplete. And it would be better to just add all custom descriptors from a specified folder and add all well-known-descriptors.

The one question is - how to get all well-known-descriptors. @jorgheymans could you, please, tell me how you implemented that? How do you get all well-known-descriptors? It would be easier and faster for me :)

**Comment by xakassi** [comment]
@jorgheymans 
Maybe you just call something like that for every well-known-type?
```
com.google.protobuf.StringValue.getDescriptor();
```

geertvb:
@xakassi The .proto files for the well-known-descriptors can be found in the com.google.protobuf:protobuf-java jar under the folder google/protobuf. You can compile these proto definitions and generate a single protobin file. We have a maven pom file to do that.
Once you have this protobin file, it can be loaded and used like any other file descriptor set.

xakassi:
Thanks, @geertvb ! I will try this approach!

xakassi:
I had some troubles with making single DescriptorSet, and I have realized it in a different way. I think, it's better in my case.

Please, guys, try your use case again with my new fixed code! @jorgheymans

geertvb:
@Xakassi Initially we also hard coded the well defined types like that but it is less flexible. Besides that we needed the protobin files for another project where we can't use that trick because we aren't using the google libraries.

xakassi:
@geertvb You can put any additional protobin files in `descriptors-folder`, Descriptors will be extracted from them and will be used for fields with `Any` type if needed.

dhakerk:
Hi we still have the same issue with your latest version here an example of the data we are treating if that may help :

here the the Timestamp is present in two location inside the Any and inside Section2Type

{
  "listIds": [
    "xxxx",
    "yyyy"
  ],
  "key": "kkkkk",
  "value": {
    "@type": "xx.yy.Type",
    "content": {
      "@type": "google.protobuf.Any",
      "type_url": "type.googleapis.com/xx.yyy.event.EventClass",
      "value": "bnbnbnbnbbnbnbnbnbbnbnbnbnbnb",
      "unpacked": {
        "@type": "xx.yyy.event.EventClass",
        "timestamp": {
          "@type": "google.protobuf.Timestamp",
          "seconds": 1610698247,
          "nanos": 339000000
        },
        "attrib1": "xxxx",
        "attrib2": "yyyy"
      }
    },
    "section2": {
      "@type": "xx.yy.Section2Type",
      "attribx": "ddddddd",
      "createdOn": {
        "@type": "google.protobuf.Timestamp",
        "seconds": 1610698248,
        "nanos": 523000000
      }
    }
  }
}

xakassi:
@dhakerk Could you, please, provide your .proto file as an example?

dhakerk:
@xakassi  nothing special in our proto definitions  here I share it with you, you may notice that the content is with type Any 

syntax = "proto3";
import "google/protobuf/any.proto";
import "xx/yyy/Section2.proto";


option java_package = "xx.yyy.proto";
// so that each message is defined in a different class
option java_multiple_files = true;

package xx.yyy;


message MainEvent {
  google.protobuf.Any content = 1;
  Section2 section2 = 2;
}
-------------------------------------------

syntax = "proto3";
import "google/protobuf/timestamp.proto";


option java_package = "xx.yyy.proto";
// so that each message is defined in a different class
option java_multiple_files = true;

package xx.yyy;

message Section2 {
  string attrib1 = 1;
  string attrib1 = 2;
  google.protobuf.Timestamp createdOn = 3;
}

Cordially

tchiotludo:
@xakassi what is the status on that ? 
To be honest, I wait this one to release a new version, need to know If you can finish it quickly or not ?

xakassi:
Hi, @tchiotludo !
Sorry, I'm really busy now, I cannot fix it quickly.

And also the error is really strange, I had no chance to reproduce it yet. It will be really great if @jorgheymans or someone from his team could provide a full example to reproduce it.

But anyway this week and the next I will be busy. So maybe you should release a new version, and the fix for Protobuf will be in the next version.

dert1714:
Hi, @tchiotludo!
Do you have any another assumtion to merge this PR except unresolved conflicts?

tchiotludo:
at @dert1714, I don't have a lot of knowledge of protobuf. 
I resolved the conflict, If you can test if it's working I could merge it

tchiotludo:
@dert1714 build failed due to incorrect unit that need to be fixed before merging

stoibazarov:
I was able to fix this error by  building proto.desc file with argument `protoc --include_imports`

Issue #534, opened 2020-12-07
Logging in with OIDC still redirects me to login page


schen2315:
Hey guys,
I setup oidc and after authenticating am brought back to the login page (login with xxxx) and can't access the actual application. When i go to /api/me i see logged is set to true, my username is there and the roles that my company org returned. Do the roles returned by my provider have to match exactly to the roles used by akhq (e.g. topic/read, topic/insert, etc. ). I am on 0.16.0 

Here's my config:
```yaml
micronaut:
  security:
    enabled: true
    oauth2:
      enabled: true
      clients:
        bsso-oauth2:
          client-id: <MY_CLIENT_ID>
          client-secret: <MY_CLIENT_SECRET>
          openid:
            issuer: <MY_PROVIDER>
security:
  default-group: no-roles
  oidc:
    enabled: true
    providers:
      bsso-oauth2:
        label: "Login with MY_PROVIDER"
        username-field: preferred_username
        groups-field: roles
        default-group: reader
```

tchiotludo:
Is this the full configuration ? 
I see that you use default group and there is an issue https://github.com/tchiotludo/akhq/pull/526 that can impact you ?

schen2315:
i only put the parts that i thought were relevant, only other part of my config is my kafka connections. Ill take a look at #526

schen2315:
i updated my config to use the suggestion made in #526 :
```yaml
security:
  default-group: no-roles
  groups:
  - name: admin
    roles:
    - topic/read
    - topic/insert
    - topic/delete
    - topic/config/update
    - node/read
    - node/config/update
    - topic/data/read
    - topic/data/insert
    - topic/data/delete
    - group/read
    - group/delete
    - group/offsets/update
    - registry/read
    - registry/insert
    - registry/update
    - registry/delete
    - registry/version/delete
    - acls/read
    - connect/read
    - connect/insert
    - connect/update
    - connect/delete
    - connect/state/update
  - name: reader
    roles:
    - topic/read
    - node/read
    - topic/data/read
    - group/read
    - registry/read
    - acls/read
    - connect/read
  - name: no-roles
    roles: []
  oidc:
    enabled: true
    providers:
      oauth2:
        label: "Login with MY_PROVIDER"
        username-field: preferred_username
        groups-field: roles
        groups:
          - name: ROLE_RETURNED_BY_PROVIDER
            groups:
              - reader
```
[...]

PR #573, opened 2021-01-28
Update  OIDC group/role mapping

hoeggi:
closes https://github.com/tchiotludo/akhq/issues/534

I think most of this are just wording issues. We have:
- akhq roles
- akhq groups
- oidc client roles

Oidc client roles are assigned to a use and should map to a corresponding group in akhq. the group in akhq then maps to its internal roles (eg: topic/reader, topic/writer...). I tried to also explain this a bit better in the readme. I only have access to keycloak so i used this as reference in my explanations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -547,11 +547,14 @@ akhq:
         google:
           label: "Login with Google"
           username-field: preferred_username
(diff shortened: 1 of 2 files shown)


tchiotludo:
Thanks ;)

PR #588, opened 2021-02-03
Adding Summary Metrics as overview for the Cluster View

theesen:
+ New endpoint returning overall stats for a cluster (partitions, replica etc.)
+ Add overview table for the cluster on top of Topics List. Showing the number of partitions, replica and insync replica per cluster
+ Add FREE NOW as kafka-hq user

### DIFF
--- a/README.md
+++ b/README.md
@@ -819,6 +819,7 @@ Documentation on Confluent 5.5 and schema references can be found [here](https:/
 * [BARMER](https://www.barmer.de/)
 * [TVG](https://www.tvg.com)
 * [Depop](https://www.depop.com)
+* [FREE NOW](https://free-now.com/)
 
 
 ## Credits
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -5,29 +5,17 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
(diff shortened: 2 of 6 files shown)


theesen:
Build fails with `ResourceNotFoundException: Connector connect-controller-test not found`. Is this related to my changes or just the build broken in dev?

tchiotludo:
The test are flakky on connect, need to find some to reproduced it :p

theesen:
@tchiotludo I have updated the routes and split them up. Do you mind having another look at it?
I am not sure when i'll have time to implement the frontend part. If you or someone else has time in between, feel free to add sth.

tchiotludo:
@theesen thanks for the work, seems to be globally good as I see. 
I don't know who can help here for the frontend ! but the backend is near the end as I see.

theesen:
@tchiotludo took me some time to get back to this issue. Before I open the actual PR there is one question on the backend that needs to be answered. I am not sure what you had in mind with `rebalancingGroups` and `emptyGroups`, how would you calculate these stats?

tchiotludo:
emptyGroups : group without any members  
image

rebalancingGroups: group where current state is Rebalancing

PR #602, opened 2021-02-14
kafka-connect-sorting-filtering

PLarboulette:
Hello, 

First PR on this repository to add a search bar in the Kafka Connect list. 

I use as an example the Schema Registry screen and implementation.

### DIFF
--- a/src/main/java/org/akhq/controllers/TopicController.java
+++ b/src/main/java/org/akhq/controllers/TopicController.java
@@ -34,6 +34,7 @@ import java.util.*;
 import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
+import org.akhq.models.Record;
 
 @Slf4j
 @Secured(Role.ROLE_TOPIC_READ)
--- a/src/test/java/org/akhq/controllers/TopicControllerTest.java
+++ b/src/test/java/org/akhq/controllers/TopicControllerTest.java
@@ -9,7 +9,7 @@ import org.akhq.utils.ResultNextList;
 import org.akhq.utils.ResultPagedList;
 import org.apache.kafka.common.config.TopicConfig;
 import org.junit.jupiter.api.*;
-
+import org.akhq.models.Record;
 import java.util.Base64;
 import java.util.List;
 import java.util.Random;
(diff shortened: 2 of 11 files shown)


src/main/java/org/akhq/controllers/TopicController.java
```diff
@@ -34,6 +34,7 @@
 import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
+import org.akhq.models.Record;
```
PLarboulette, line 37: I have compilation problems here, with ambiguous usage of Record (java.lang.Record and Record of package models). I import it explicitly.

src/test/java/org/akhq/controllers/TopicControllerTest.java
```diff
@@ -9,7 +9,7 @@
 import org.akhq.utils.ResultPagedList;
 import org.apache.kafka.common.config.TopicConfig;
 import org.junit.jupiter.api.*;
-
+import org.akhq.models.Record;
```
PLarboulette, line 12: I have compilation problems here, with ambiguous usage of Record (java.lang.Record and Record of package models). I import it explicitly.

PLarboulette:
I have some problems with the "GroupControllerTest", function listApi() line 42 (expected: <1> but was: <5>) 
I can't understand why, I think I don't touch this part and this fail doesn't seem to be a problem for the global build.

tchiotludo:
@PLarboulette : some test are flaky ;)
Last one works as you see !
Thanks for this one, will look at this soon ;)

tchiotludo (commented):
Hi @PLarboulette 
Sorry for late response, really busy here ! 
I have made some test; the search is working fine :+1: 
But the pagination seems to be not working, the total page is not displayed, and changing to next page just keep the same result than previous one. 

Can you look at this ?

tchiotludo (changes requested):
See comment above

PLarboulette:
Hello, 

Oh sorry, It's pushed with a fix, I think it's better. 

FYI, during my work in Kafka Connect, I find a similar to mine : https://github.com/tchiotludo/akhq/pull/622

Issue #611, opened 2021-02-25
Increase test coverage with github workflow and a "verified" testsuite


akramer-zibra:
Hej!
I like your project, so I am sad test coverage got lost..
and therefore I wanna share this idea with you:

Let's configure a `test` github workflow and start initially with only "verified" testcode.
"verified" testcode could be tests which already succeed.

...it is difficult to fix broken testcode without knowing the original intention.
Maybe failing testcode is already outdated or will drop out by changes or refactorings anyway. 

This kind of a "verified" testsuite could be a starting point to ensure future changes will not have any bad side effects.

JUnit 5 supports testing subsets via `@Tag`-annotations.
It is as easy as annotate code with `@Tag("verified")` .

The command for a suitable github action could look like this: `./gradlew :test -PincludeTags='verified'`

If you think this idea is it worth I'll gladly create a Pull Request for this.

Cheers! :)

tchiotludo:
Hi @akramer-zibra, sorry for late response. 
I have read you, but to be honest, I don't understand the goal of this PR ? 

All test from akhq is working, just some are flaky (since it was integration test, sometime kafka connect is not ready, so the test failed). 

If a test really don't work, it will be removed like that [this test for example](https://github.com/tchiotludo/akhq/blob/dev/src/test/java/org/akhq/controllers/AclsControllerTest.java#L19). 

So I don't understand why a verified tag ? 
All test is working, and I couldn't release AKHQ if the test is not working. 

Maybe I don't understand the goal of this issue / PR ?

akramer-zibra:
Hi @tchiotludo .
locally on my development machine 63 of 133 tests fail...

![grafik](https://user-images.githubusercontent.com/5008216/110322518-863fe380-8013-11eb-985e-1f5c272a6005.png)

This in mind I created this issue. Hmm.
...strange that there is such a difference between local testing and ci testing results.

I haven't had a look at the github ci pipeline yet. D'oh.

So this issue is cleared.
Thanks! :)

PR #612, opened 2021-02-25
Github workflow for testing and a "verified" testsuite 

akramer-zibra:
Implementation for #611 

Starts "verified" testsuite with some succeeding tests.


akramer-zibra:
Request cleared. Details described in #611

Issue #614, opened 2021-03-03
Enabling the magnifier to see log errors when connect/read is enabled


fabiofilz:
Hi,

I enabled the connect/read role, but the magnifying glass icon in the connector screen is not showing so the user with read-only access cannot see the connector's parameters, tasks or the error logs.

Basically the same as connect/update, but not enabling the value fields and update button.

Would it be possible?

Thank you.
Fabio

tchiotludo:
Seems to be possible :) 
Can you share screnshot to be sure to understand ?

fabiofilz:
Hi @tchiotludo,

Yes, sure!

Using connect/read role

![image](https://user-images.githubusercontent.com/45266582/109859002-63e54900-7c54-11eb-9262-6a86897f5070.png)


Using connect/update role
![image](https://user-images.githubusercontent.com/45266582/109856187-1d421f80-7c51-11eb-9584-de36bd1bcb65.png)

Tasks
![image](https://user-images.githubusercontent.com/45266582/109856347-4c589100-7c51-11eb-944c-ef69067431a8.png)

This is the Update button
![image](https://user-images.githubusercontent.com/45266582/109856448-685c3280-7c51-11eb-83d1-4ae23a8d012a.png)

Thank you.
Fabio

tchiotludo:
Ok will be simple :) Thanks :)

tchiotludo:
You can try on dev version 
fix by #620

fabiofilz:
Hi @tchiotludo,

Could you please create a new TAG with the changes, please?

I checked the Dockerhub, but I didn't see a new TAG. The 0.0.17 was created a month ago.

Thank you.
Fabio

![image](https://user-images.githubusercontent.com/45266582/111539021-51dfcc00-8765-11eb-8666-1ab3a4d59571.png)

tchiotludo:
this one :smile: : 
![image](https://user-images.githubusercontent.com/2064609/111539214-e5b59600-876d-11eb-862a-2b763b4a888e.png)
the `dev` version ;)

fabiofilz:
Hi @tchiotludo,

Sorry, I thought it was a final release with the changes.

I have just deployed the DEV tag and I am using the role connect/read.

I am still not able to see the magnifier to access the error logs and I also notice that when I change between one Kafka Connect and the other it is not refreshing the page (list of connectors). It was not happening in the previous version.

[...]

PR #620, opened 2021-03-08
fix connect update permissions

alozano3:
Hi, I want to allow a user to modify the state of a connector/task (resume, pause, restart) and not allow him to update the configs of the connector.

The first requirement is covered by the role `connect/state/update` and the second requirement by the role `connect/update`

The problem is that if I don't assign the role `connect/update` to the user, the search icon that redirects to the tasks tab where you can pause/resume is hidden.
connectList

Here is the tab that cannot be accessed without the `connect/update` role.
tasks

My suggestion is that the role `connect/update` only should apply when trying to update the configs (making the button disappear) and allow users with the role `connect/state/update` to change the state. 
configs

### DIFF
--- a/client/src/containers/Connect/ConnectDetail/ConnectConfigs/ConnectConfigs.jsx
+++ b/client/src/containers/Connect/ConnectDetail/ConnectConfigs/ConnectConfigs.jsx
@@ -139,6 +139,7 @@ class ConnectConfigs extends Form {
     let title = '';
     let { formData } = this.state;
     const errors = [];
+    const roles = this.state.roles || {};
     const errorMessage = this.validateProperty({ name: plugin.name, value: formData[plugin.name] });
     if (errorMessage) {
       errors[plugin.name] = errorMessage;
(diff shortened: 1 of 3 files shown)


tchiotludo:
Hi @alozano3, it will fix this issue also I think : https://github.com/tchiotludo/akhq/issues/614, no ?

alozano3:
Yes! I didn't realize that there already was an opened issue. I read the #[614](https://github.com/tchiotludo/akhq/issues/614) and the code of this PR doesn't disable the inputs of the config tabs, although the update button is hidden. Do you think it's worth to disable all the input fields?

tchiotludo:
It can be more explicit to disabled input, people will understand clearly that no action is possible If you can

alozano3:
I disabled the inputs and the transforms textarea if the role `connect/update` is not granted.

alozano3:
@tchiotludo did you have time to review the last change?

tchiotludo:
Perfect thanks :)

PR #623, opened 2021-03-08
FIX-pagination-schema-registry 

PLarboulette:
Hello 

During my work on the Kafka connect part, I can see there is also a bug on the Schema Registry screen.

The totalPage doesn't appear and we can continue on the next page even if it doesn't exist (and Java doesn't seem to like that) 

image

image

I think this code will fix the problem


tchiotludo:
also perfect, thanks :+1:

Issue #484, opened 2020-10-30
Usage of a groups-filter-regexp to limit consumer group view


wilkej:
Hello everyone and thank you for this project.

I was wondering if it is possible to have a filter-regexp for the consumer group view. 

I have a limit view on the topics configured but If I go to the consumer group view I see all groups with topic names. Yes, I can't look into the topics but I know additional ones. 

I think it would be nice to limit the view on the consumer groups or show only the consumer groups which consume the topics I'm allowed to view.

Thank you for your feedback

tchiotludo:
Hi, 
It's not possible for now, but PR are welcome :+1: 
I just merged the same PR for Connect here : https://github.com/tchiotludo/akhq/pull/477 if it can help implementing this features

PR #628, opened 2021-03-09
Filter consumer groups regexp

alozano3:
Hi, this PR adds a new feature to filter the consumer groups by regexp, as it is done with topics or connects. I also added a test class for the consumer groups component.

This PR also fix the issue #484.


tchiotludo:
work perfectly, thanks a lot :tada:

PR #629, opened 2021-03-09
describe only filtered topics

alozano3:
Hi, this code aims to improve the efficiency of the requests that retrieve topics information.
The old behaviour was to call `kafkaWrapper.describeTopics` and `kafkaWrapper.describeTopicsOffsets` methods and then filter the results with topic regex and then add the filtered results to the UI.

The proposed logic is to first filter the topic list with the regexp, and then call the mtehods `kafkaWrapper.describeTopics` and `kafkaWrapper.describeTopicsOffsets`. That way, several requests are saved, and at the same time, I consider it is more secure because information of topics not available is not requested.


tchiotludo:
Nice catch, thanks :+1:

PR #631, opened 2021-03-10
add schema registry ssl example

alozano3:
Add schema registry ssl configuration example

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,7 +258,9 @@ akhq:
         type: "confluent"
         basic-auth-username: avnadmin
         basic-auth-password: {{password}}
-        properties: {}
+        properties:
+          schema.registry.ssl.truststore.location: {{path}}/avnadmin.truststore.jks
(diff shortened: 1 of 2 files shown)


alozano3:
Hi, it seems that there is a flaky test. When the tclass est ConsumerGroupRepositoryTest executes before the test class ConnectRepositoryTest everything is OK. The problem is when it executes after, because the class ConnectRepositoryTest creates several consumer groups and the cluster is reused in the next test class. I tried to find a way to clean those consumer groups in the @Before or even create a new cluster, but I couldn't reorganize the code without impacting all the classes. 
Do you have any suggestion?

tchiotludo:
Yes there is flakky test, I know. 
I don't think it's related to order of test, I've the feeling that there is a kind of timeout on connect api that will failed all the others one but maybe i'm wrong. 
In general, the flakky test are only 1 times, I retry and forgot ...

PR #634, opened 2021-03-13
refacto-pagination

PLarboulette:
Hello, 

Just a bit of refacto on the Pagination component and functions

### DIFF
--- a/client/src/utils/pagination.js
+++ b/client/src/utils/pagination.js
@@ -1,0 +1,16 @@
+
+export const handlePageChange = ({ currentTarget: input, state }) => {
+    const { value } = input;
+    state.pageNumber = value;
+};
+
+export const getPageNumber = (value, totalPageNumber) => {
+    if (value <= 0) {
+        value = 1;
+    } else if (value > totalPageNumber) {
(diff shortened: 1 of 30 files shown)


client/src/utils/pagination.js
```diff
@@ -0,0 +1,16 @@
+
+export const handlePageChange = ({ currentTarget: input, state }) => {
```
PLarboulette, line 2: I am not sure but I think this function is never called, I don't find the way to use it

PLarboulette: The goal of the function is only to update the state.pageNumber, this is is already done by the  handlePageChangeSubmission function

`this.setState({ pageNumber: pageNumber }, () => {
// Callback 
}`

tchiotludo:
Hi @PLarboulette ! 
Thanks for this ! 
Maybe I can ask you another thing ? The pagination don't change the url and I found this annoying, Can you handle this case with this PR ? :crossed_fingers:

PLarboulette:
Hey, 

Yes, I will look that ASAP 👍

Issue #593, opened 2021-02-08
Login page does not appear with basic auth and no-roles config


alozano3:
Hi, I am trying to set basic authentication with **_no-roles_** as the default group. The Login page does not appear and it seems it gets stuck in a loop. When I set the default-group to _admin_ , it does not ask for credentials and directly shows the dashboard.

Config:
```yaml
    # Auth & Roles (optional)
    security:
      default-group: no-roles # Default groups for all the user even unlogged user
      basic-auth:
        - username: admin
          password: XXX # Password in sha256
          groups:
            - admin
```

This is the error that I am seeing int the browser console.
![image](https://user-images.githubusercontent.com/31214345/107212280-02400f00-6a07-11eb-9531-7b2398d16cdb.png)

tchiotludo:
What version of AKHQ are you using ? 
Can you provide a full application.yml example ?

alozano3:
Hi, I  already fix the problem. I migrated from 0.14.0 to 0.17.0 and I had to add the below conf:

```yaml
  micronaut:
    security:
      enabled: true
```

xakassi:
Hi, @tchiotludo !
I'm also migrating now from 0.14.1 to 0.17.0 and spent a lot of time trying to understand why UI does not work with basic-auth.

Maybe it worth to add this to all security configuration examples in README since it is mandatory now?
```
micronaut:
    security:
      enabled: true
```

tchiotludo:
@xakassi PR are welcome :smile: :+1:

xakassi:
I've prepared a PR. @tchiotludo , please check it :)

PR #639, opened 2021-03-17
Minor fix for README

xakassi:
Regarding #593
Add info about mandatory `micronaut.security.enabled: true`.
Also fix some minor typos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@
   - Create a topic
   - Configure a topic
   - Delete a topic
-- **Browse Topic datas**
+- **Browse Topic data**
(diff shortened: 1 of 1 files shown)


xakassi:
I have noticed that `org.akhq.controllers.ConnectControllerTest` unit test often fails, and I need to force push to make it run again, and it finish successfully only at a specific time of a day :)
Is it ok?

tchiotludo:
the test is flakky I know, ... and I don't know how to fix that :sob: 
Just retry seems to work without no specific time ...

tchiotludo:
Thanks a lot ;)

Issue #646, opened 2021-03-25
/akhq/templates/deployment.yaml


Hyunwiii:
{{- with .Values.image.annotations }}

should be 

{{- with .Values.annotations }}

Thanks.

tchiotludo:
Seems that there is an issue here ! PR are welcome :+1:

PR #651, opened 2021-04-01
Update deployment.yaml

Hyunwiii:
Issue: /akhq/templates/deployment.yaml #646


tchiotludo:
Hey, 

Look at here : https://github.com/tchiotludo/akhq/blob/594cda9a82e64028ff34780706740f98c3ea8054/helm/akhq/templates/deployment.yaml#L6-L8
There is already an usage for annoations, maybe rename ` .Values.image.annotations` to ` .Values.podAnnotations` instead ?

PR #659, opened 2021-04-07
Allows List or String in topics-filter-regex

### LINKED ISSUE #656: topicsFilterRegexp doesn't accept List
Hello,
In order to simplify a long regex with many "or" elements, I looked in the code and figured it was possibly allowing a List
 https://github.com/tchiotludo/akhq/blob/594cda9a82e64028ff34780706740f98c3ea8054/src/main/java/org/akhq/repositories/TopicRepository.java#L171-L178

Such as that the following config : 
````yaml
security:
    groups:
      reader-all:
        name: "reader-all"
        roles:
        - "topic/read"
        - "topic/data/read"
        attributes:
          topics-filter-regexp:
          - "project1.*"
          - "project2.*"
````
would work.
But in fact it's failing during Topic filtering in the repository and when you look at the JWT, this is what we have (list in list)
````yaml
"topicsFilterRegexp": [
    [
      "project1.*",
      "project2.*"
    ]
  ]
````

You specifically check for "instance of List" so is this a bug ? 

Would it make sense to change the configuration from this : 
````yaml
topics-filter-regexp: "^(priv_bsmjlea_|pub_transport-unit|pub_loading|priv_sfr_|quality_result|fabrication|consumption).*"
````
to this ?
````yaml
topics-filter-regexp:
- priv_bsmjlea_
- pub_transport-unit
- pub_loading
- priv_sfr_
- quality_result
- fabrication
- consumption
````

Obviously applies to topics, connects and group regex.

Happy to contribute if you're fine with this change.

twobeeb:
Closes #656 
I haven't been able to test with security enabled but I tested with default group using both List and String yaml configuration, and it works as intended.
All thanks to Micronaut that maps a String property into a List<String> seamlessly (creating a list of one). 

See attached test.
I didn't test connects-filter-regx or groups-filter-regex since the code is 100% reused.

@tchiotludo 
Could you test with security enabled to confirm that the generated JWT looks ok (compared to #656) ?
If not possible, I'll find some time next week to test on a real cluster.


tchiotludo:
Seems to be perfect, thanks :+1:

Issue #662, opened 2021-04-08
add curl to the dockerfile


survivant:
I had issues with the probes
````
 Warning  Unhealthy  2m16s (x6 over 3m36s)  kubelet            Liveness probe failed: dial tcp 192.168.158.111:8081: connect: connection refused
  Normal   Killing    2m16s (x2 over 3m16s)  kubelet            Container akhq failed liveness probe, will be restarted
  Warning  Unhealthy  2m12s (x8 over 3m42s)  kubelet            Readiness probe failed: Get "http://192.168.158.111:8081/health": dial tcp 192.168.158.111:8081: connect: connection refused
````

so I wanted to connect into the pod to try to call manually the URL so see what was the values returned, but **wget** and **curl**  are not available in the dockerfile.

and we don't have root access, so I couldn't installed it.

for now, I just put in comments the probes and it works fine.

**Comment by tchiotludo** [comment]
Sound a nice addition, just drop a PR on the docker file and I will merge ! 
(just for information I change the probe endpoint port on last version, and the helm charts reflect this change, in case you are using an old akhq version)

**Comment by vutkin** [comment]
Hey @tchiotludo , this is very bad decision, curl adds a lot of vulnerabilities for free. It brings more problem than solves.

<img width="1141" alt="image" src="https://user-images.githubusercontent.com/25753488/178788429-13ea4250-63ca-4d9d-9f2a-9707685ba35d.png">
 Thats just about curl, and there are a lot of dependencies that curl uses.

@survivant I advise you to modify pod and add another container with curl for debug purposes:

```
1. Add a sidecar container:
cat <<EOF > patch-file.yaml
spec:
  template:
    spec:
      shareProcessNamespace: true
      securityContext:
        runAsUser: 0
        fsGroup: 0
      containers:
      - name: akhq
      - name: debug
        image: alpine/curl
        command: ["/bin/sh"]
        args: ["-c", "tail -f /dev/null"]
        securityContext:
          capabilities:
            add:
            - SYS_PTRACE
```
kubectl patch deployment <prefix>-connect --patch "$(cat patch-file.yaml)"

PR #663, opened 2021-04-08
added curl to the image

survivant:
for https://github.com/tchiotludo/akhq/issues/662


survivant:
@tchiotludo here the PR :)

Issue #680, opened 2021-04-21
Missing password function in Connect Config UI


hoeggi:
If you try to view the configuration of a connector like this:
`/ui/{kafka-cluster-name}/connect/{connector-name}/definition/jdbc-db2-source/configs` (sample jdbc connector)
you can see and edit the configuration. 

We are also running some custom connectors right now and viewing the config fails with following error in the chrome dev tool console:
```
react-dom.production.min.js:209 TypeError: j.a.password is not a function
    at a.e.handleDefinition (ConnectConfigs.jsx:108)
    at ConnectConfigs.jsx:70
    at Array.forEach (<anonymous>)
    at a.e.handleSchema (ConnectConfigs.jsx:68)
    at a.e.renderForm (ConnectConfigs.jsx:330)
    at a.<anonymous> (ConnectConfigs.jsx:56)
    at fo (react-dom.production.min.js:131)
    at os (react-dom.production.min.js:212)
    at hc (react-dom.production.min.js:255)
    at t.unstable_runWithPriority (scheduler.production.min.js:19)
es @ react-dom.production.min.js:209
ConnectConfigs.jsx:108 Uncaught (in promise) TypeError: j.a.password is not a function
    at a.e.handleDefinition (ConnectConfigs.jsx:108)
    at ConnectConfigs.jsx:70
    at Array.forEach (<anonymous>)
    at a.e.handleSchema (ConnectConfigs.jsx:68)
    at a.e.renderForm (ConnectConfigs.jsx:330)
    at a.<anonymous> (ConnectConfigs.jsx:56)
    at fo (react-dom.production.min.js:131)
    at os (react-dom.production.min.js:212)
    at hc (react-dom.production.min.js:255)
    at t.unstable_runWithPriority (scheduler.production.min.js:19)
```

This happens if the configuration has the field of type `password` and is set to `required`, the default jdbc connector has the password field non required and shows the config.

Difference here is that all required fields are being validated and I think somewhere in the validation code tree shaking or some obfuscation removes the validation function.

tchiotludo:
fix in #681

PR #681, opened 2021-04-22
Use Joi.string().required() for passwords

thomaspeklak:
As far as I have seen there is no method `Joi.password`. For passwords it makes sense that you treat them as a simple string.

Solves #680


tchiotludo:
Thanks :+1:

Issue #697, opened 2021-05-12
Log failed authentications and authorizations


gquintana:
For security reasons, I need to be able to track failed login attempts and access denied errors.

The first parts seems easy (I will try to do a PR) https://micronaut-projects.github.io/micronaut-security/latest/guide/#securityEvents
The second part is already implemented through access logs which contain user name

PR #699, opened 2021-05-13
feat(auth): log login failures

gquintana:
Try to implement #697 

The result is log saying:
```
2021-05-13 22:03:17,607 WARN  pGroup-1-3 u.LoginFailedEventListener Login failed reason USER_NOT_FOUND, username unknown, message User Not Found
2021-05-13 22:04:39,488 WARN  pGroup-1-4 u.LoginFailedEventListener Login failed reason CREDENTIALS_DO_NOT_MATCH, username unknown, message Credentials Do Not Match
```
Which is not very satisfactory, because the username is always unknown.

The AuthenticationFailed class contains:
```
    @Override
    public Optional<UserDetails> getUserDetails() {
        return Optional.empty();
    }
```
So I think it requires a change in Micronaut in AuthenticationFailed and in various AuthenticationProviders


gquintana:
Build fails due to SchemaControllerTest#listApi test which also fails on dev branch

PR #702, opened 2021-05-15
Fix secrets value from map to string

NiloofarShahbaz:
secrets must be string so that it can be converted to base64 in secrets.yaml


tchiotludo:
Good catch, thanks ;)

PR #709, opened 2021-05-21
Show size in bytes in the /topic page #616

### LINKED ISSUE #616: Show size in bytes in the /topic page
Is there a configuration i can use to show the topic size in bytes in the /topic page.
The size column stays at a unit until the value for that unit changes which does not give the right picture.

Example:
Entity: Apple
Entity Size: 500 KB
Size displayed for 2000 Apples = 1GB
Size displayed for 3999 Apples = 1GB
Size displayed for 4000 Apples = 2GB

ThiagoTeodoro:
Implements a function to show bytes size with a round way.
With this commit now its possible se values likce 1.2567 GB

Resolve #616

### DIFF
--- a/client/src/containers/Topic/TopicList/TopicList.jsx
+++ b/client/src/containers/Topic/TopicList/TopicList.jsx
@@ -181,7 +181,7 @@ class TopicList extends Root {
         name: topic.name,
         count: topic.size,
         lastWrite: undefined,
-        size: showBytes(topic.logDirSize, 0),
+        size: showBytes(topic.logDirSize),
         partitionsTotal: topic.partitions.length,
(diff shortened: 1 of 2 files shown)


tchiotludo (changes requested):
image

image

Sorry for late reply, lot of works this times. 
Just test the PR and I don't understand why 4 characters after dot ? why not just 3 ? 
Maybe use a lib like : https://www.npmjs.com/package/pretty-bytes to display it ?

ThiagoTeodoro:
@tchiotludo i changed for 3 characters after dot.
I don't think we need a lib for this, you have a couple scenarios to use show bytes convert and considering that, it's gonna have more performance with a single function in the client.

PR #735, opened 2021-06-14
Changed condition to display Login Form

twobeeb:
Login Form must be visible when LDAP is enabled within Micronaut instead of "LDAP to Akhq mapping" class


tchiotludo:
too quick ;) 
```
org.opentest4j.AssertionFailedError: Login Form must not be active if only OIDC is enabled ==> expected: <false> but was: <true>
````

PR #746, opened 2021-06-29
fix(helm): Use toYaml Go template for configmap

AnatomicJC:


tchiotludo:
agree with that :) 
It wil be better but we need to be consistent, can you make the change everywhere (secrets, ...) ?

AnatomicJC:
Hi @tchiotludo 

I updated the `values.yaml` file accordingly.

I also replaced `indent` with `nindent`. Do you see other things to update ?

tchiotludo:
`configuration` is also yml and can't be use as yaml in values

AnatomicJC:
I reverted my last commit about `indent` and `nindent`
I tested on a local cluster and configmap seems ok now. What do you think ?

tchiotludo:
perfect, thanks :)

AnatomicJC:
Thanks to you for this project :+1:

PR #780, opened 2021-08-03
Require key for produce onto compacted topic

### LINKED ISSUE #133: Exception when producing to  compacted topic without a key
```
[Producer clientId=producer-1] Got error produce response with correlation id 217 on topic-partition json-0, retrying (2147483432 attempts left). Error: CORRUPT_MESSAGE
```
- Make key mandatory
- See if you could have a better management of retry (2147483432 seems very long :smile:)

timtebeek:
Fixes #133 @tchiotludo

### DIFF
--- a/src/main/java/org/akhq/models/Topic.java
+++ b/src/main/java/org/akhq/models/Topic.java
@@ -126,15 +126,13 @@ public class Topic {
             return false;
         }
 
-        return isCompacted(clusterId, this.getName(), configRepository);
+        return isCompacted(configRepository.findByTopic(clusterId, this.getName()));
     }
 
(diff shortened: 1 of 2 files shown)


tchiotludo:
Just a concern here. 
If the ACL doesn't allow you to read the config, it will failed with exception. 
On this case, I think it's better to remove the check

timtebeek:
Not sure if I read your comments correctly, but I've added a try-catch-log block in case there's any `ExecutionException` while checking to see if topic is compacted. If you had intended any other/further changes please let me know.

And thanks for the other couple of merges as well. Nice to see these added! :)

PR #791, opened 2021-08-14
Add multi-message produce feature

xpr3sso:
Hi @tchiotludo,

This feature allows the user to input multiple records as key-value pairs at once as one-liners like:
```
key1:{"value": 41}
key2:{"value": 42}
```
or
```
key1_{"value": 41};key2_{"value": 42}
```

This can help when dealing with larger sets of generated data - there is a limit for the request size, though :D  
Multiple key-value separators can be chosen from, to support various naming conventions people might have.

Looks like this on the front end:
image
The "Key" field gets disabled when "Multiple messages" is selected, the multi-message fields for standard "Single message" producing. Placeholders for both modes are added to give some guidance.

What do you say? Could this be useful in general?

### DIFF
--- a/client/src/components/Form/Form.jsx
+++ b/client/src/components/Form/Form.jsx
@@ -112,8 +112,9 @@ class Form extends Root {
     );
   };
 
-  renderJSONInput = (name, label, onChange) => {
+  renderJSONInput = (name, label, onChange, textMode, options) => {
     const { formData, errors } = this.state;
+    const inputMode = textMode ? "text" : (formData.schemaType === "PROTOBUF" ? "protobuf"  : "json")
     return (
       <div className="form-group row">
         {label !== '' ? (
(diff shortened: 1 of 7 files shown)


tchiotludo:
Hi :) 
I think it's a cool features ! Thanks for this one :) 
My first impression (without a lot of reflexion) : 
Maybe we can use a standard method for this multiple message (in order to handle key ? 
Maybe : 
```
key1,{"id": "123"}
key2,{"id": "456"}
```
or 
```json
{"key": "key1","value": {"id": "123"}}
{"key": "key2","value": {"id": "456"}}
```
A simple checkbox for multiple message, and we display a notice about the format needed ? 

xpr3sso:
Good idea! The checkbox combined with a text field for the separator would do the job. Let me check and get back to you :)

xpr3sso:
Alright, was just able to try some layout variants. This is how the result looks like:
image
Looks way cleaner and key-value pairs can be split with any char(-combination) the user provides ;)

tchiotludo:
@xpr3sso : does this one can be review ?

xpr3sso:
Hi @tchiotludo,
yes, should be ready for review now :)  No quite sure why some tests are failing now, though.

tchiotludo:
thanks @xpr3sso works great :+1:

PR #794, opened 2021-08-16
feat(auth): align header authentication user/group settings with LDAP and OIDC

piotrp:
Note: this changes behavior of `akhq.security.header-auth.users` to make it work the same as in OIDC/LDAP. Please check whether what I did is correct.

I haven't tested it yet, will comment here when I verify it works in my environment. I'm creating this PR to have tests run and get feedback if what I did missed something.


piotrp:
Verified, works in my environment

piotrp:
This build failure seems unrelated?

tchiotludo:
Some test are flakky. 

Thanks for all the contribution :+1:

Issue #719, opened 2021-06-02
Query parameters don't survive a login


chiodonia:
We deployed AKHQ with LDAP integration and we are using deep linking with query params to jump i.e. into a topic.
However, if you are not previously logged in to AKHQ you are redirected to the login page. Once you log in you are then redirected to the main AKHQ page and initial query params are lost.

Would it be possible to make query parameters survive the login?

tchiotludo:
I think it's possible ;)

grosswilerp:
When will this issue be fixed?

tchiotludo:
When someone (you) will make a PR :smile:

grosswilerp:
> When someone (you maybe ?) will make a PR

@tchiotludo I have created a PR as requested. Let me know if I need to make any changes.

PR #797, opened 2021-08-20
feat(auth): added support for redirecting to initial page after login #719

grosswilerp:
This PR should implement the feature request from issue #719.

When checking if the user has a session, before redirecting to `/ui/login` it will check if the URL is different than `/ui/login` and stores it in `sessionStorage.setItem("returnTo")`.

Once the user has logged in, it will check if there is a value in `sessionStorage.setItem("returnTo")` and sets this instead of the default `/ui` for the `pathname`.

### DIFF
--- a/client/src/components/Root/Root.jsx
+++ b/client/src/components/Root/Root.jsx
@@ -1,4 +1,4 @@
-import {Component} from 'react';
+import { Component } from 'react';
 import axios from 'axios';
 import { get, put, post, remove } from '../../utils/api';
 
@@ -7,6 +7,12 @@ class Root extends Component {
   cancel = axios.CancelToken.source();
 
   componentWillUnmount() {
+    const pathname = window.location.pathname;
+
+    if (pathname !== '/ui/login') {
+      sessionStorage.setItem('returnTo', pathname + (window.location.search || ''));
+    }
+
     this.cancelAxiosRequests();
   }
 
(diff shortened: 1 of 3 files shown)


tchiotludo:
Just make a quick try and it don't seems to work. 

Adding here : https://github.com/tchiotludo/akhq/blob/dev/client/src/components/Root/Root.jsx#L9
```js
  componentWillUnmount() {
    const pathname = window.location.pathname;

    if (pathname !== '/ui/login') {
      sessionStorage.setItem ('returnTo', pathname+window.location.search);
    }

    this.cancelAxiosRequests();
  }
```
and removing from `Routes.js` seems to be a better place to handle it ? 

trancee:
I am not sure this is a better way. Since the redirection back to the initial page should only happen if the user was not logged in and needs to log in. That's why I had put it in `Routes.js` where the logic is to redirect to the login page.

If we add it to `componentWillUnmount` it will be called every time a component is unmounted. I am not sure that is the intended behavior?

I have tested my code before creating the PR, and it worked for me. Unless I did something wrong when testing, could you share how you tested it, so I can reproduce this on my end as well?

tchiotludo:
The expected behaviour is to redirect to previous page on every login. 
I think you cover the hard redirect when there is no acls with `no-roles` group. 

The `componentWillUnmount` try to cover a case when the user have a login button on the page and some acls defined (like `reader` for default, and logged with `admin`). I expected for them to get redirected to previous page. 

What to do you think about that ?

grosswilerp:
> The expected behaviour is to redirect to previous page on every login.
> I think you cover the hard redirect when there is no acls with `no-roles` group.

Yes, you are absolutely right there.

> The `componentWillUnmount` try to cover a case when the user have a login button on the page and some acls defined (like `reader` for default, and logged with `admin`). I expected for them to get redirected to previous page.
> 
> What to do you think about that ?

That makes sense. I will adjust my PR accordingly. Thank you for the explanation.

grosswilerp:
@tchiotludo Did you check the PR after your suggestions? Is there anything else I need to do?

tchiotludo:
Oups totally forgot this one ;) 
All is good, thanks :+1:

PR #803, opened 2021-09-06
Fix: Parsing timestamps fails when producing messages

xarx00:
The commit fixes two related problems that prevent producing messages if AVRO contains a timestamp field. Production of such messages wasn't possible at all in AKHQ version 0.18.0, but the major problems have been fixed since then. However, some problems still remain.

Let's have an AVRO like this:
```
{
  "type": "record",
  "name": "myrecord",
  "fields": [
    {
      "name": "cas",
      "type": {
        "type": "long",
        "logicalType": "timestamp-millis"
      }
    }
  ]
}
```

1. The message 
```
{
  "cas": 12345
}
```
fails to be produced because Integer cannot be converted to Instant.

2. The message 
```
{
  "cas": "2021-09-05T20:31:53.582+02"
}
```
cannot be produced because the parser cannot parse short time-zones.

This patch fixes both problems. Timestamp strings without time-zone can still be parsed, Even negative numbers (Integers and Longs) can be parsed.

### DIFF
--- a/src/main/java/org/akhq/utils/AvroSerializer.java
+++ b/src/main/java/org/akhq/utils/AvroSerializer.java
@@ -13,14 +13,15 @@ import java.math.BigDecimal;
 import java.math.MathContext;
 import java.math.RoundingMode;
 import java.nio.ByteBuffer;
-import java.time.*;
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalTime;
 import java.time.format.DateTimeFormatter;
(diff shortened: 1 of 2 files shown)


tchiotludo:
Thanks @xarx00 , maybe you can add a unit test to validate the behavior please ?

xarx00:
I added unit tests for short form of TZ offset. AvroSerializer methods like timestampMillisSerializer() need testing too, but I would have to change their visibility (they are private), which I didn't dare to do.

PR #815, opened 2021-09-15
add option to create a service account and specify a default user id in securityContext

AndreZiviani:
I had two problems trying to use `avhq` in AWS EKS:
- Lack of `ServiceAccount` in the helm chart
- The container was running with the `akhq` user (uid=1000 gid=1000) but it wasn't specified in the `securityContext` so AWS created the token file with strict permissions (only the root user could read it)

This PR adds an option to automatically create the `ServiceAccount` and specify some options in the `securitContext` by default

### DIFF
--- a/helm/akhq/Chart.yaml
+++ b/helm/akhq/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.18.0"
 description: Kafka GUI for Apache Kafka to manage topics, topics data, consumers group, schema registry, connect and more...
 name: akhq
-version: 0.2.2
+version: 0.2.3
 keywords:
   - kafka
   - confluent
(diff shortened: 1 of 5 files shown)


tchiotludo (changes requested):
Can you use the standard way to handle service for a better readability please ? 

```yaml 
serviceAccountName: {{ include "bla.serviceAccountName" . }}

{{/*
Create the name of the service account to use
*/}}
{{- define "bla.serviceAccountName" -}}
{{- if .Values.serviceAccount.create }}
{{- default (include "bla.fullname" .) .Values.serviceAccount.name }}
{{- else }}
{{- default "default" .Values.serviceAccount.name }}
{{- end }}
{{- end }}
```
if the user input `serviceAccountName` & `serviceAccountCreate` the yaml will be invalid

AndreZiviani:
@tchiotludo can you please review the changes?

Issue #833, opened 2021-09-29
Decode Avro messages without schema id


bgranvea:
I have a topic containing Avro messages without schema metadata. I would like to be able to display them in akhq but I can't modify the producer. All messages in this topic follow the same schema.

As far as I can tell, this is not possible today to configure a topic=>schema id mapping (or maybe topic => schema without using schema registry at all).

This feature was discussed in #181 but finally the need was for serialization and deserialization was never implemented. Is that right?

tchiotludo:
Yes it's right, this was never done for now. 
The whole logic is [here](https://github.com/tchiotludo/akhq/blob/dev/src/main/java/org/akhq/models/Record.java) and specially [here](https://github.com/tchiotludo/akhq/blob/dev/src/main/java/org/akhq/models/Record.java#L186), if no schema is found, you can try to detect that message is an avro, and convert it to `GenericAvro` and call : 
```java
                GenericRecord record = (GenericRecord) toType;
                return avroToJsonSerializer.toJson(record);
```
 in order to have a json and it's done :smile: 

a PR is welcome on that

bgranvea:
> Yes it's right, this was never done for now. The whole logic is [here](https://github.com/tchiotludo/akhq/blob/dev/src/main/java/org/akhq/models/Record.java) and specially [here](https://github.com/tchiotludo/akhq/blob/dev/src/main/java/org/akhq/models/Record.java#L186), if no schema is found, you can try to detect that message is an avro, and convert it to `GenericAvro` and call :

just to be sure that I have understood correctly: I'm not an Avro specialist but from what I understand, I need to have a schema to decode the message. 

My idea was to configure some topic=>schema mappings, just like `ProtobufToJsonDeserializer`. In `Record`, if no schemaId is embedded in message, I look at these mappings and if I find a match, I try to decode the message with this schema.

tchiotludo:
ahh ok :smile: 
I don't understand, I think the schema was embed in the message (a normal avro is a binary file that contains in header the schema, and in the body the data). 
In your case, the schema was not in the message ? are you sure ? 
If not, the mapping is only a duplication off the worker of protobuf effectively

PR #843, opened 2021-10-05
Custom deserializer for avro raw format

bgranvea:
This PR implements #833. 

It allows the decoding of Avro raw binary format, with a mapping configuration similar to what already exists for Protobuf, for example:

```
        akhq:
          connections:
            my-cluster:
              deserialization:
                avro-raw:
                  schemas-folder: "/app/avro_schemas"
                  topics-mapping:
                    - topic-regex: "data.*"
                      value-schema-file: "Schema.avsc"
```

A couple of remarks regarding this PR:
- it only addresses Avro "raw" binary format, that is objects encoded with low-level `DatumWriter` (no header), which is my use case. It could be extended to support single-object encoding (https://avro.apache.org/docs/current/spec.html#single_object_encoding) and it could also be interesting to decode object container files (https://avro.apache.org/docs/current/spec.html#Object+Container+Files) which embed the schema
- there is no support for schema referencing another schemas
- I'm not a Gradle specialist but I didn't manage to have Avro schemas used for unit tests that are compiled by Gradle Avro plugin as well as also packaged as resources, so I ended up duplicating the files...
- I modified `AvroDeserializer` so that displayed JSON keeps the same field order than Avro schema. This is convenient for unit tests and I think it is also for the user

### DIFF
--- a/src/main/java/org/akhq/configs/AvroTopicsMapping.java
+++ b/src/main/java/org/akhq/configs/AvroTopicsMapping.java
@@ -1,0 +1,14 @@
+package org.akhq.configs;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class AvroTopicsMapping {
+    String topicRegex;
+    String keySchemaFile;
+    String valueSchemaFile;
+}
--- a/src/main/java/org/akhq/configs/Connection.java
+++ b/src/main/java/org/akhq/configs/Connection.java
@@ -18,7 +18,7 @@ import java.util.Map;
 public class Connection extends AbstractProperties {
     SchemaRegistry schemaRegistry;
     List<Connect> connect;
-    ProtobufDeserializationTopicsMapping deserialization;
+    Deserialization deserialization;
     UiOptions uiOptions = new UiOptions();
 
     public Connection(@Parameter String name) {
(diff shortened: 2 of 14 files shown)


bgranvea:
Sorry for that, it was difficult to setup my dev environment, unit tests don't seem to work at all on Windows (some weird Kafka errors), I finally managed to get it work with WSL.

I fixed most of the unit tests, but there is still one failing test. I don't understand because I have the same error when I switch to the dev branch so this doesn't seem to be related to my PR. Any idea?

```
  org.akhq.controllers.SchemaControllerTest

    ✔ deleteNotExistApi()
    ✔ crud()
    ✘ listApi()

      org.opentest4j.AssertionFailedError: expected: <5> but was: <3>
          at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
          at org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
          at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
          at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
          at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:510)
          at org.akhq.controllers.SchemaControllerTest.listApi(SchemaControllerTest.java:35)
```

I also noticed this in Record:

```
    @JsonIgnore
    private SchemaRegistryClient client;

    private ProtobufToJsonDeserializer protobufToJsonDeserializer;
```

Shouldn't this field also be annotated with `JsonIgnore`?

PR #856, opened 2021-10-15
Display __consumer_offsets contents

### LINKED ISSUE #654: Handle special topic like Schema Registry and __consumer_offsets and display data
Hello, 

I think I find a strange behavior in the topic list records. When I want to read the __consumer_offsets topic, the consumer tries to use the Schema Registry to deserialize key and values (probably because both is in a binary format ?) 

This behavior results in many calls on the Registry, with 404 response because the id used to get the schema doesn't exist (obviously) 

I think I can fix it, it's just to know if you are already anything in progress about this :) 

Have a nice day !

timtebeek:
Fixes #654

### DIFF
--- a/src/main/java/org/akhq/models/Record.java
+++ b/src/main/java/org/akhq/models/Record.java
@@ -192,9 +192,25 @@ public class Record {
             }
         } else if (topic.isInternalTopic() && topic.getName().equals("__consumer_offsets")) {
             if (isKey) {
-                return GroupMetadataManager.readMessageKey(ByteBuffer.wrap(payload)).toString();
+                try {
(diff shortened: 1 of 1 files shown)


tchiotludo:
also added __transaction_state topic since we are loading the class needed :=)

timtebeek:
Hmm; the returned `OffsetAndMetadata` class does not expose any getters for the JSON serializer to use, so you'd see an empty Object. We could:
1. Leave it as is
2. Map OffsetAndMetadata fields into a Map
3. Map OffsetAndMetadata to a DTO with getters

I'd lean towards option 1 or 2, especially since this is all quite exceptional handling for special topics.

PR #899, opened 2021-11-02
Add Vodeno as akhq user

karolcienkosz:
As kafka administrator at Vodeno I admit that we use akhq and we love it! I'm adding us as an official user.


tchiotludo:
Thanks :+1:

PR #905, opened 2021-11-12
fix: Take into account default topic configuration during topic creation

### LINKED ISSUE #821: Topics options are not taken into account
Using AKHQ v 0.18.0.

It seems that topics options are not taken into account.

Here is my configuration:

```yaml
akhq:
  topic:
    partition: 12
    replication: 3
    retention: 1209600000
```

But in the UI, I still got default values when creating a new topic.

Am I missing something ?

Thank for your support,

  Cédric

Thomas-Dangleterre:
Fixes #821.

I created it as a draft because I wanna know if it's okay for you @tchiotludo before carrying on (adding tests...)

Regards


tchiotludo:
Hi, 
Seems to be ok for myself.
Please just add some unit test on the new endpoint and rename the route to `("api/topic/defaults-configs")`
Thanks 👍

Thomas-Dangleterre:
Done :) 

Thanks

Issue #810, opened 2021-09-10
IAM MSK support


tooptoop4:
Support for https://docs.aws.amazon.com/msk/latest/developerguide/iam-access-control.html

tchiotludo:
For aws part, I don't have any AWS account, so this one will need a PR for people using it.

tooptoop4:
already works.

application.yml:
```
micronaut:
  server:
    port: 8089
    host: "0.0.0.0"
akhq:
  connections:
    my-cluster-sasl:
      properties:
        bootstrap.servers: "redact:9098"
        security.protocol: SASL_SSL
        sasl.mechanism: AWS_MSK_IAM
        sasl.jaas.config: software.amazon.msk.auth.iam.IAMLoginModule required;
        sasl.client.callback.handler.class: software.amazon.msk.auth.iam.IAMClientCallbackHandler
        ssl.truststore.location: /home/user/kafka.client.truststore.jks
        ssl.truststore.password: changeit
        ssl.keystore.location: /home/user/kafka.client.truststore.jks
        ssl.keystore.password: changeit
      schema-registry:
        url: "http://localhost:8081"
```

```
java -Dmicronaut.config.files=application.yml -cp /home/user/iamjar/aws-msk-iam-auth-1.1.1-all.jar:/home/user/iamjar/akhq.jar org.akhq.App
```

tchiotludo:
Reopening for docs and maybe to include the jar on main package 🤔

PR #948, opened 2021-12-14
Add a dependency library to enable IAM access control for MSK(#810)

moremagic:
Support for https://docs.aws.amazon.com/msk/latest/developerguide/iam-access-control.html

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -124,6 +124,9 @@ dependencies {
 
     // client
     implementation project(":client")
+
+    //AWS MSK IAM Auth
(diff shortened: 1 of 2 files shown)


tchiotludo:
Just add the dependency allow to connect to msk ?
If yes, can you add an example of configuration in the doc please ?

moremagic:
Yes. We are just adding a dependency.

I'll try to fix the documentation later.
Please wait a moment.

moremagic:
Sorry for the delay.
I added the documentation.

PR #958, opened 2021-12-16
Replaced Map<String, Object> with AKHQClaimResponseAttributes

twobeeb:
Checklist:
- [ ] Dev
- [ ] Tests
- [ ] Doc


tchiotludo:
@twobeeb can you look at these : https://github.com/tchiotludo/akhq/pull/961 ? 
I think it's a better solution

PR #959, opened 2021-12-16
Test for RestApiClaimProvider

twobeeb:
This test should fail on dev branch


twobeeb:
I disabled KafkaWrapperFilter for the scope of this test as it conflicts with this mocked service

PR #961, opened 2021-12-20
fix/external

tchiotludo:
- Test for RestApiClaimProvider
- Disable KafkaWrapperFilter which interracts incorrectly with RestApiExternalService Controller
- wip


twobeeb (approved):
Yes, it looks better this way. 
I think I originally wanted to leave room for evolution by keeping ``Map<String, Object> attributes`` but now I think it's better to strong-type everything.

Issue #835, opened 2021-09-29
IndexOutOfBoundsException thrown in topics view when changing to cluster with fewer pages


ahalaj:
Using version 0.18.0

When I am on Topics view in the UI and I go to the last page and then I change my cluster to a different one, which has fewer pages, I get this exception:

```
java.lang.IndexOutOfBoundsException: toIndex = 150
	at java.base/java.util.AbstractList.subListRangeCheck(Unknown Source)
	at java.base/java.util.ArrayList.subList(Unknown Source)
	at org.akhq.utils.PagedList.of(PagedList.java:67)
	at org.akhq.repositories.TopicRepository.list(TopicRepository.java:62)
	at org.akhq.controllers.TopicController.list(TopicController.java:81)
	at org.akhq.controllers.$TopicControllerDefinition$$exec1.invokeInternal(Unknown Source)
	at io.micronaut.context.AbstractExecutableMethod.invoke(AbstractExecutableMethod.java:151)
	at io.micronaut.context.DefaultBeanContext$4.invoke(DefaultBeanContext.java:482)
	at io.micronaut.web.router.AbstractRouteMatch.execute(AbstractRouteMatch.java:303)
	at io.micronaut.web.router.RouteMatch.execute(RouteMatch.java:121)
	at io.micronaut.http.server.netty.RoutingInBoundHandler.emitRouteResponse(RoutingInBoundHandler.java:1533)
	at io.micronaut.http.server.netty.RoutingInBoundHandler.access$2100(RoutingInBoundHandler.java:151)
	at io.micronaut.http.server.netty.RoutingInBoundHandler$6$1.request(RoutingInBoundHandler.java:1500)
	at io.reactivex.internal.operators.flowable.FlowableFlatMap$InnerSubscriber.onSubscribe(FlowableFlatMap.java:656)
	at io.micronaut.http.server.netty.RoutingInBoundHandler$6.doSubscribe(RoutingInBoundHandler.java:1488)
	at io.micronaut.http.server.netty.RoutingInBoundHandler$6.lambda$subscribe$0(RoutingInBoundHandler.java:1482)
	at io.micrometer.core.instrument.composite.CompositeTimer.record(CompositeTimer.java:79)
	at io.micrometer.core.instrument.Timer.lambda$wrap$0(Timer.java:157)
	at io.micronaut.scheduling.instrument.InvocationInstrumenterWrappedRunnable.run(InvocationInstrumenterWrappedRunnable.java:47)
[...]

PR #962, opened 2021-12-22
Fix: IndexOutOfBoundsException thrown in topics view when changing to…

EdwinFajardoBarrera:
… cluster with fewer pages.

Closes issue #835

When the cluster is changed now it forces the state to have pageNumber equals to 1 to avoid failure if the new cluster has fewer pages than the old one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@
 * [Decathlon](https://www.decathlon.fr/)
 * [Depop](https://www.depop.com)
 * [GetYourGuide](https://www.getyourguide.com)
+* [Kitopi](https://kitopi.com)
 * [Klarna](https://www.klarna.com)
 * [La Redoute](https://laredoute.io/)
 * [Leroy Merlin](https://www.leroymerlin.fr/)
--- a/build.gradle
+++ b/build.gradle
@@ -124,6 +124,9 @@ dependencies {
 
     // client
     implementation project(":client")
+
+    //AWS MSK IAM Auth
+    implementation group: 'software.amazon.msk', name: 'aws-msk-iam-auth', version: '1.1.1'
 }
 
 /**********************************************************************************************************************\
--- a/client/src/containers/Connect/ConnectList/ConnectList.jsx
+++ b/client/src/containers/Connect/ConnectList/ConnectList.jsx
@@ -49,7 +49,7 @@ class ConnectList extends Root {
     const query =  new URLSearchParams(this.props.location.search);
     this.setState({
       searchData: { search: (query.get('search'))? query.get('search') : searchData.search },
-      pageNumber: (query.get('page'))? parseInt(query.get('page')) : parseInt(pageNumber)
+      pageNumber: (query.get('page')) ? parseInt(query.get('page')) : parseInt(pageNumber)
     }, () => {
       this.getConnectDefinitions();
     });
@@ -60,7 +60,9 @@ class ConnectList extends Root {
       this.cancelAxiosRequests();
       this.renewCancelToken();
 
-      this.componentDidMount();
+      this.setState({ pageNumber: 1 }, () => {
+        this.componentDidMount();
+      });
     }
   }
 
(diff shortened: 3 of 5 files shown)


EdwinFajardoBarrera:
@tchiotludo If you can review my PR I would appreciate it, I await your comments, greetings!

tchiotludo:
@EdwinFajardoBarrera it seems to be a good fix !
Is it possible to handle this on list please ?

EdwinFajardoBarrera:
> @EdwinFajardoBarrera it seems to be a good fix ! Is it possible to handle this on list please ?

Sorry I don't get you, how can I help you?

tchiotludo:
For example this [one](https://github.com/tchiotludo/akhq/blob/dev/client/src/containers/Connect/ConnectList/ConnectList.jsx#L52), I've the feeling the bug is also here ?
May you have a look on all pages with pagination ?

EdwinFajardoBarrera:
> 

For sure, take it for granted

EdwinFajardoBarrera:
@tchiotludo I'm having issues configuring multiple clusters, actually, this is working

- [ ]  Kafka-0
- [ ]  Schema-registry-0
- [ ]  Connector-0 ** Docker container stops
- [ ]  Kafka-1
- [ ]  Schema-registry-1
- [ ]  Connector-1

I leave my docker-compose-multiple-cluster.yml file: https://codeshare.io/1YbNvX

My application-dev.yml connections are as follows:

```
akhq:
  connections:
    local-0:
      properties:
        bootstrap.servers: "localhost:9092"
      schema-registry:
        url: "http://localhost:8085"
      connect:
      - name: "connect-0"
        url: "http://localhost:8083"
    local-1:
      properties:
        bootstrap.servers: "localhost:9072"
      schema-registry:
        url: "http://localhost:8075"
      connect:
      - name: "connect-1"
        url: "http://localhost:8084"

```

Thanks for your help

tchiotludo:
can you give the log of the failed connect please ? 
I don't see any reasoon like that

EdwinFajardoBarrera:
> can you give the log of the failed connect please ? I don't see any reasoon like that

I just got this:

image

image

I see that the first connect instance that I select makes the other shutdown.

```
[2021-12-22 19:16:37,849] INFO [Worker clientId=connect-1, groupId=kafka-connect-1] Session key updated (org.apache.kafka.connect.runtime.distributed.DistributedHerder)

Dec 22, 2021 7:16:35 PM org.glassfish.jersey.internal.inject.Providers checkProviderRuntime

WARNING: A provider org.apache.kafka.connect.runtime.rest.resources.RootResource registered in SERVER runtime does not implement any provider interfaces applicable in the SERVER runtime. Due to constraint configuration problems the provider org.apache.kafka.connect.runtime.rest.resources.RootResource will be ignored.
```

tchiotludo:
You have a mistake corrected here : https://codeshare.io/1YbNvX
But I'm not really sure you're setup is possible like that (on the same docker host / multiple kafka, zookepper), I've the feeling they will try to make a unique cluster

EdwinFajardoBarrera:
> For example this [one](https://github.com/tchiotludo/akhq/blob/dev/client/src/containers/Connect/ConnectList/ConnectList.jsx#L52), I've the feeling the bug is also here ? May you have a look on all pages with pagination ?

I just checked and fixed [this](https://github.com/tchiotludo/akhq/blob/dev/client/src/containers/Connect/ConnectList/ConnectList.jsx#L52) and apparently, there is no other page (where using pagination) that replicates this bug.

tchiotludo:
THanks for having a deeper look :+1:

EdwinFajardoBarrera:
> THanks for having a deeper look 👍

No problem, let me know if there is another issue with priority (If this is in the backend is better for me)

tchiotludo:
https://github.com/tchiotludo/akhq/labels/backend you will find the backend issue, feel free to pick up one that sound interesting for you

PR #963, opened 2021-12-22
add Kitopi as akhq user

maciej-wilczynski:
At Kitopi we are using akhq from the beginning, and we are happy to share that information.


tchiotludo:
Thanks :+1:

PR #964, opened 2021-12-22
New ACL model

twobeeb:
#897 

- [x] Refactor ACL model
- [ ] Define the Resource/Role matrix
- [ ] Add all default permissions in application.yml
- [ ] Cleanup old model and references
- [x] Refactor authentication / ClaimProvider / LDAP / OIDC / Header / Basic
- [x] Manage unauthenticated users (micronaut.security.enabled=false)
- [x] Integrate with Topics Controller
- [ ] Integrate with  Groups Controller
- [ ] Integrate with  Node Controller
- [ ] Integrate with  Connects Controller
- [ ] Integrate with  into Schemas Controller
- [ ] Modify UI (...)


tchiotludo:
for the controller part we need to had a custom micronaut validation that replaces the `@Secured` from micronaut

twobeeb:
> for the controller part we need to had a custom micronaut validation that replaces the `@Secured` from micronaut

I agree for something like `@MySecured(resource="TOPIC" operation="CONSUME") for example, but how do you pass the current topic to the annotation ?

tchiotludo:
> > for the controller part we need to had a custom micronaut validation that replaces the `@Secured` from micronaut
> 
> I agree for something like `@MySecured(resource="TOPIC" operation="CONSUME")` for example, but how do you pass the current topic to the annotation ?

Will be the tricky part here ! 

Here is an example from a private repository : 

```java 
@Target({ElementType.METHOD, ElementType.TYPE})
@Retention(RetentionPolicy.RUNTIME)
@Inherited
@Documented
public @interface HasAnyPermission {
    Permission value();
}


@Singleton
public class HasAnyPermissionSecurityRule extends AbstractSecurityRule {
    @Inject
    public HasAnyPermissionSecurityRule(RolesFinder rolesFinder) {
        super(rolesFinder);
    }

    @Override
    public Publisher<SecurityRuleResult> check(HttpRequest<?> request, @Nullable RouteMatch<?> routeMatch, @Nullable Authentication authentication) {
        if (!(routeMatch instanceof MethodBasedRouteMatch)) {
            return Flowable.just(SecurityRuleResult.UNKNOWN);
        }

        MethodBasedRouteMatch<?, ?> methodRoute = ((MethodBasedRouteMatch<?, ?>) routeMatch);
        if (!methodRoute.hasAnnotation(HasAnyPermission.class)) {
            return Flowable.just(SecurityRuleResult.UNKNOWN);
        }

        Optional<Permission> hasAnyPermission = methodRoute.getValue(HasAnyPermission.class, Permission.class);
        if (hasAnyPermission.isEmpty()) {
            return Flowable.just(SecurityRuleResult.UNKNOWN);
        }

        if (this.getUser(request).isEmpty()) {
            return Flowable.just(SecurityRuleResult.REJECTED);
        }

        if (aclService.hasAny(this.getUser(request).get(), hasAnyPermission.get())) {
            return Flowable.just(SecurityRuleResult.ALLOWED);
        } else {
            return Flowable.just(SecurityRuleResult.REJECTED);
        }
    }
}
```

`routeMatch` will have a lot of useful method so I've no doubt the the cluster with be available on route args.
For the topic name, consumer group name, ... we will not add a second validation on the method body.
The annotation will be here to be sure the user has a right on the reources, operation for this cluster only I think

tchiotludo:
We will have to this `isUserAllowedRequest` but the annotation is a another security, on the body method we will need to control only the patterns, all the other control will be on the annotation. 
From a security point, we will be sure that the most important (and most used control) will be done no matter the body content.

twobeeb:
Are you aligned with this direction https://github.com/tchiotludo/akhq/pull/964/commits/36dde2bec7c68389c203d00f00584929078bfd47 ?

twobeeb:
@tchiotludo Everything builds correctly. You can only test using the API.
The new behavior is like this : 
- All endpoints are secured through bindings, even when anonymous.
- If you need AKHQ with no authent and full admin the config must be : 
  ````yaml
  micronaut.security.enabled: false
  akhq:
    security:
      global-bindings: 
        - admin # refers to a default binding with ALL permissions
  ````
- If you need AKHQ with no authent and read only the config must be : 
   ````yaml
  micronaut.security.enabled: false
  akhq:
    security:
      global-bindings: 
        - reader # refers to a default binding with all READ-ONLY permissions
  ````
- If you need AKHQ with Authentication and user/group mappings : 
   ````yaml
  micronaut.security.enabled: true
  akhq:
    security:
      global-bindings: [ ] # Empty list
    basic-auth:
      - username: user
        password: d74ff0ee8da3b9806b18c877dbf29bbde50b5bd8e4dad7a3a725000feb82e8f1
        bindings:
          - reader-project1
      - username: admin
        password: d74ff0ee8da3b9806b18c877dbf29bbde50b5bd8e4dad7a3a725000feb82e8f1
        bindings:
          - admin
  ````

grosswilerp:
I am looking forward for this PR to be merged into AKHQ. What is the plan for this? @tchiotludo @twobeeb

tchiotludo:
@grosswilerp, this PR cannot be merged because it's a WIP and a lot of feature is missing. 
As I understand, @twobeeb can't work anymore on the subject.
So someone need to restart the work and finalize (I don't have the time also myself)

grosswilerp:
> @grosswilerp, this PR cannot be merged because it's a WIP and a lot of feature is missing. As I understand, @twobeeb can't work anymore on the subject. So someone need to restart the work and finalize (I don't have the time also myself)

Thank you for your reply. When you say a lot of features are missing, what do you mean exactly?
I would like to continue the work rather than starting over, unless it is completely out of date.

Can you give me a quick update on what's still missing or not working?

tchiotludo:
@grosswilerp not really was is done so far and I think the PR is still usable, the core of AKHQ don't have a lot of change on ACLS part since this one. 

As I understand, the checklist is providing you what is missing: 
- Finalize the backend (implement on every controller) 
- Finalize the unit test that is failing for now
- After, start the implementation on the front end (where nothing as started)
- The final polish can be done by me (doc, configuration examples, cleanup, ...) 

Are you able to handle this? front & back?

grosswilerp:
> As I understand, the checklist is providing you what is missing:
> 
> * Finalize the backend (implement on every controller)
> * Finalize the unit test that is failing for now
> * After, start the implementation on the front end (where nothing as started)
> * The final polish can be done by me (doc, configuration examples, cleanup, ...)
> 
> Are you able to handle this? front & back?

I do not have any experience with Micronaut so far, but I will look into it. As for the remaining open points, I think I might be able to contribute on those, not sure what has to be done on the frontend side exactly.

tchiotludo:
On the front, nothing is done has I understand.

tchiotludo:
close in favor of #1472, this one serve as inspiration

Issue #564, opened 2021-01-15
Displaying large messages in topics is stalling the UI


rlamcaiso:
Hello @tchiotludo.   Thank you for your tremendous effort with this project.  We are just starting to use the application in our Kafka development and it has been a big help.  

An issue that we are encountering is that the app would hang when we view a topic with many large messages.  Would it be possible to disable the message preview window, and provide a button for individual messages to view an abbreviated string of the message (ie. 250 characters) and with a download button to download the full message?

Thank you so much for your time.

Regards,
Rick

tchiotludo:
What size is your "with many large messages"  ? 
250 is really to low so I need to understand where we need to put the limit (backend or frontend) !

rlamcaiso:
Thank you for the prompt response.   Our bulk messages range from 2MB to 60MB, and there could be hundreds of these messages generated per day and published to the topic.

tchiotludo:
ok so I your case, the need will be server side ! 
60 mb will always froze the brower :smile:

rlamcaiso:
Yes.  We don't know how else to manage this performance issue with our current payloads.  I appreciate your help!

EdwinFajardoBarrera:
Hello @tchiotludo, I would like to work on a solution to this, can u assign it to me?

tchiotludo:
How will you handle this ?

EdwinFajardoBarrera:
> How will you handle this ?

Well, I think the backend is required to catch large messages and only send a small preview. I see the limit is almost 10MB but if we send 10MB messages the problem could just persist, so maybe I think 5MB could be ok. What do you think?

tchiotludo:
It's a good solution, please add a configuration options with a decent default value (5MB)

EdwinFajardoBarrera:
Hello @tchiotludo, I still working on a solution to this issue. I'm catching the size of messages in the backend and the problem stills if there are many messages of length near to the limit (4.9MB for example) and I think cutting the message does not give a good UX because JSON format fails and maybe is not worth seeing only a cut part of a message. I'm thinking about only showing a message informing this situation and making more visible the "share" option to see the full message.

Something like this:

![image](https://user-images.githubusercontent.com/36389489/147698009-c8bc1563-a977-4960-aedd-9a6578538d4b.png)

What do you think?

tchiotludo:
I still think the truncate is the better option than nothing, debugging without any message is impossible, I prefer a snippet 
[...]

PR #970, opened 2021-12-29
enhacement: prevent UI freezing with large messages and single messages downloading

EdwinFajardoBarrera:
Closes: #564

Changes:

- Backend: 
I added a const (**maxKakfaMessageLength**) to limit the size (1000000 bytes) of the messages of a topic when retrieving a list of all those, in order to prevent UI freezing.
In order to set the new value of the message, I added a setter method in the Record class.

I also added a function (**filterMessageLength**) that is called after filter records that will be responded as the request of the frontend where is checked if the limit of the message is exceeded and then is sliced in a short preview.

- UI:
I added a new action to the Table component which calls the **onDownload** event, in order to allow download a message when previewing a single one. I also implemented the method which downloads a file with the message value as content (.json format).

image
image

### DIFF
--- a/src/main/java/org/akhq/models/Record.java
+++ b/src/main/java/org/akhq/models/Record.java
@@ -79,6 +79,9 @@ public class Record {
 
     private byte MAGIC_BYTE;
 
+    @Setter(AccessLevel.NONE)
+    private Boolean truncated;
+
     public Record(RecordMetadata record, SchemaRegistryType schemaRegistryType, byte[] bytesKey, byte[] bytesValue, Map<String, String> headers, Topic topic) {
         this.MAGIC_BYTE = schemaRegistryType.getMagicByte();
         this.topic = topic;
@@ -90,6 +93,7 @@ public class Record {
         this.bytesValue = bytesValue;
         this.valueSchemaId = getAvroSchemaId(this.bytesValue);
         this.headers = headers;
+        this.truncated = false;
     }
 
     public Record(SchemaRegistryClient client, ConsumerRecord<byte[], byte[]> record, SchemaRegistryType schemaRegistryType, Deserializer kafkaAvroDeserializer,
@@ -120,6 +124,7 @@ public class Record {
         this.kafkaProtoDeserializer = kafkaProtoDeserializer;
         this.avroToJsonSerializer = avroToJsonSerializer;
         this.kafkaJsonDeserializer = kafkaJsonDeserializer;
+        this.truncated = false;
     }
 
     public String getKey() {
@@ -149,6 +154,10 @@ public class Record {
 
     public void setValue(String value) {
         this.value = value;
+    }
+
+    public void setTruncated(Boolean truncated) {
+        this.truncated = truncated;
     }
 
     private String convertToString(byte[] payload, Integer schemaId, boolean isKey) {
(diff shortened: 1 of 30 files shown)


EdwinFajardoBarrera:
@tchiotludo what do you think about this approach?

src/main/java/org/akhq/repositories/RecordRepository.java
```diff
@@ -1268,5 +1273,14 @@ public TailOptions(String clusterId, List<String> topics) {
         private final long end;
         private final KafkaConsumer<byte[], byte[]> consumer;
     }
+
+    private void filterMessageLength(Record record) {
```
tchiotludo, line 1277: I would prefer add a property `truncated` as bool on record to avoid to have translation on backend

src/main/java/org/akhq/repositories/RecordRepository.java
```diff
@@ -76,6 +77,8 @@
     @Value("${akhq.clients-defaults.consumer.properties.max.poll.records:50}")
     protected int maxPollRecords;
 
+    private final int maxKafkaMessageLength = 1000000; // 1MB
```
tchiotludo, line 80: Please add `@Value` to be able to configure the limit and add a default on `application.yml`

tchiotludo (changes requested):
Hi @EdwinFajardoBarrera, 

Sorry I was sure that I respond to you :smile: 

This seems to be good overall, just minor changes

src/main/java/org/akhq/repositories/RecordRepository.java
EdwinFajardoBarrera: Do you refer to sending the entire message and only catching this property in front end?

tchiotludo: you truncate the message on the backend plus add a property to tell the message is truncated. 
The front will only add the ... too large message ... if truncated is true

EdwinFajardoBarrera:
I made the changes @tchiotludo, let me know if there is something else to change

src/main/java/org/akhq/models/Record.java
```diff
@@ -71,12 +72,16 @@
...
 
     @Getter(AccessLevel.NONE)
+    @Setter(AccessLevel.NONE)
     private String value;
 
     private final List<String> exceptions = new ArrayList<>();
 
     private byte MAGIC_BYTE;
 
+    @Setter(AccessLevel.NONE)
```
tchiotludo, line 82: You need to remove this one, other else the value will never be sent on the ui part

src/main/java/org/akhq/repositories/RecordRepository.java
```diff
@@ -76,6 +77,9 @@
     @Value("${akhq.clients-defaults.consumer.properties.max.poll.records:50}")
     protected int maxPollRecords;
 
+    @Value("${akhq.kafka-max-message-length}")
```
tchiotludo, line 80: use `akhq.topic-data.kafka-max-message-length` and please add some docs [here](https://github.com/tchiotludo/akhq/blob/dev/docs/docs/configuration/akhq.md) and add the value [here](https://github.com/tchiotludo/akhq/blob/dev/application.example.yml)

tchiotludo (changes requested):

PR #995, opened 2022-01-26
Link consumergroup topics offset to data single

### LINKED ISSUE #992: Link from consumer group topics offset column to topic message
When viewing consumer group topics it would be nice to link from the offset column directly to the associated message. So from:
`/ui/{cluster}/group/{consumergroup}/topics` to:
`/ui/{cluster}/topic/{topic}/data?single=true&partition={partition}&offset={offset}`.

That way it's easier to verify what is the last message that a consumer with lag has not yet processed.
Just a nice little touch to help link the two pages together, since they are different views on the same data.
Possibly not even that hard to do, but I've not yet looked in to this as I found it during the workday. :)

timtebeek:
Fixes #992 

Have very little experience with React & jsx , so this should be considered a rough first attempt.

Also was not yet able to verify locally, as [the development server](https://akhq.io/docs/dev.html#development-server) environment does not have any consumer groups by default; and I didn't yet find the time to create a topic & consumer. Might be nice to expand the development environment with [the test data seen here](https://github.com/tchiotludo/akhq/blob/dev/docker-compose.yml#L102).

### DIFF
--- a/client/src/containers/ConsumerGroup/ConsumerGroupDetail/ConsumerGroupTopics/ConsumerGroupTopics.jsx
+++ b/client/src/containers/ConsumerGroup/ConsumerGroupDetail/ConsumerGroupTopics/ConsumerGroupTopics.jsx
@@ -90,7 +90,14 @@ class ConsumerGroupTopics extends Root {
               colName: 'Offset',
               type: 'text',
               cell: obj => {
-                return this.handleOptional(obj.offset);
+                if (obj.offset !== undefined && obj.offset !== '') {
+                  return (
+                    <Link to={`/ui/${this.state.selectedCluster}/topic/${obj.name}/data?single=true&partition=${obj.partition}&offset=${obj.offset}`}>
+                      {obj.offset}
+                    </Link>
+                  );
+                }
+                return <label>-</label>;
               }
             },
             {


tchiotludo:
It's work well :) nice first jsx :+1:

timtebeek:
Awesome, thanks for checking and merging it so quickly!

tchiotludo:
You have a lot of chance, I was blocking few hours this night on akhq :sweat_smile:

timtebeek:
Tried it just now, and I'm wondering if we should link to offset-1 instead. :sweat_smile: 
Getting a few not-found responses, whereas the offset-1 record is present.

tchiotludo:
You are true, we must even more don't go to single page also.
The better will be to go offset-1 on this partition and let the full page (will all the record).
You can know if there is a transaction (that led -1) or not (that to offset). 
The better for the user is so to display a full page :)

timtebeek:
Something like this then? https://github.com/tchiotludo/akhq/compare/dev...timtebeek:link_to_offset-1?expand=1
Wasn't sure what is and is not allowed within the `${...}` expressions.

tchiotludo:
I don't remember for the `${...}` but I think it's good 
and the link must be : 
```jsx
<Link to={`/ui/${this.state.selectedCluster}/topic/${obj.name}/data?partition=${obj.partition}&after=${obj.partition}-${obj.offset - 1}`}>
```

Issue #983, opened 2022-01-18
Add extra information to nodes page - controller status, partition load


jgrammen-agilitypr:
Any chance the following extra information could be added to the nodes page?

![image](https://user-images.githubusercontent.com/22965146/149945486-4326e6d2-4343-4300-b2ac-d6f6b9a8221f.png)

Having the controller status and the partition load (% of total), is very helpful to see when the cluster has stabilized post reboot.

jgrammen-agilitypr:
maybe the node page could also indicate when a node is missing:
![image](https://user-images.githubusercontent.com/22965146/149945971-5689c1b1-dea6-4200-b167-490c39027cc5.png)

tchiotludo:
PR are welcome on that, 
- 1st is a simple ui. 
- 2nd seems to be more complicated (I don't know if the api return the valueà

jgrammen-agilitypr:
so i see you have merged the controller status in. I will next tackle the partition information, now @tchiotludo has clarified that it needs back end updates.

jgrammen-agilitypr:
I am struggling a little bit to figure out where in the backend java code I need to insert the code to fetch the partition information so that it can be included in the info passed to the /nodes page. Can anyone provide some guidance?

jgrammen-agilitypr:
After some more looking through the code, it is apparent that somehow the NodeController needs access to the topic repository, but it is un clear to me how this can happen
any advice?

tchiotludo:
You need to inject topic repository : 
```java
@Inject
TopicRepository topicRepository
```
and add some filtering on partitions based on node id

jgrammen-agilitypr:
There is now a pull request to add the additional "partitions" column. Code is functional as far as I can test (I don't know how to join a second node to the Kafka cluster in docker). Comments are welcome, especially given this is my first time contributing Java code.

PR #999, opened 2022-01-27
#983 add node controller status to nodes page

jgrammen-agilitypr:
First attempt at adding additional information to the nodes page. Namely the controller status and the partitions information.


tchiotludo:
I have to remove the `Number of partitions` information, there is not this values on the api. 
If you want it, we will need to update the backend also

PR #1000, opened 2022-01-27
Uncheck all topics in consumer group update view by default.

dezka:
Currently when resetting offsets for a consumer group, all of the topics it subscribes to will be checked for the reset operation. This could be dangerous for a user that isn't paying attention and only trying to reset offsets for a particular topic. I propose we have all topics unchecked by default.

I also noticed that if you don't have any topics checked, and click `Update`, the toast notification will show a success despite it not actually finishing behind the scenes. Adding a notification to give more clarity here.


tchiotludo:
Thanks, just correct the checked that is wrong

dezka:
@tchiotludo thank you. When do you think a `0.21.0` release will happen to include this change?

tchiotludo:
Not really soon, there is some hardwork in progress

PR #1002, opened 2022-01-27
Link to offset 1 from ConsumerGroupTopic

timtebeek:
As discussed in https://github.com/tchiotludo/akhq/pull/995#issuecomment-1023576643


tchiotludo:
I added sort=oldest that is needed :) Thanks

PR #1003, opened 2022-01-28
Update application.example.yml

wktsugue:
Fixed YAML error on line 277


wktsugue:
The standalone execution command line was not working on PowerShell. Fixed by using double-quotes.

PR #1015, opened 2022-02-08
Handle logical types local-timestamp-millis and local-timestamp-micros

opeti:
Avro specification defines additional logical types, which are not yet supported by akhq ([Avro specification](https://avro.apache.org/docs/current/spec.html#Logical+Types)):
- local-timestamp-millis
- local-timestamp-micros

When using schema with logical types of the above, akhq fails to convert the messages to JSON and visualize them.

This PR adds support for handling the mentioned logical types.

### DIFF
--- a/src/main/java/org/akhq/utils/AvroDeserializer.java
+++ b/src/main/java/org/akhq/utils/AvroDeserializer.java
@@ -14,6 +14,7 @@ import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.time.Instant;
 import java.time.LocalDate;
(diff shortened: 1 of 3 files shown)


tchiotludo:
Perfect! Thanks :+1:

lostiniceland:
@tchiotludo Would it be possible to cut a release? 0.20 is already 6 months old and we would need exactly this fix. Thanks

tchiotludo:
it's possible: https://github.com/tchiotludo/akhq/releases/tag/0.21.0 :smile:

PR #1043, opened 2022-03-17
feat/fix: Toggle "time ago" and timestamp with interactive tooltip

### LINKED ISSUE #938: Allow to toggle "time ago" and timestamp
After adding [feat(ui): display time ago on every datetime](https://github.com/tchiotludo/akhq/commit/d7343161362459d7fc196d5dbceda93ca0ef9daa) all timestamps are only available as mouse over labels. Therefore, it is no longer possible to copy a timestamp from the ui.

It would be great if this behaviour would be configurable: either the timestamp is displayed and a time ago label or vice versa.

RomanLamsal1337:
Fixes the current behavior of not being able to select the ISO date displayed in the tooltips of the TopicData table (see comment below). 

Also added a dropdown to switch between displaying relative times via `react-timeago` or plain ISO dates. The tooltip remains, but switches to either option.

fixes #938

### DIFF
--- a/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
+++ b/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
@@ -67,7 +67,7 @@ class TopicData extends Root {
     percent: 0,
     loading: true,
     canDownload: false,
-    dateFormat: constants.SETTINGS_VALUES.TOPIC_DATA.DATE_FORMAT.RELATIVE
+    dateTimeFormat: constants.SETTINGS_VALUES.TOPIC_DATA.DATE_TIME_FORMAT.RELATIVE
   };
 
   searchFilterTypes = [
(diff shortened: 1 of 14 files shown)


client/src/containers/Topic/Topic/TopicData/TopicData.jsx
```diff
@@ -910,7 +928,13 @@ class TopicData extends Root {
                     colName: 'Date',
                     type: 'text',
                     cell: (obj, col) => {
-                      return (<TimeAgo date={Date.parse(obj[col.accessor])} title={obj[col.accessor]}/>);
+                      const isoDate = obj[col.accessor]
+                      const TimeAgoComp = <TimeAgo date={Date.parse(isoDate)} title={""}/>
+                      return (
+                          <Tooltip arrow title={!this.state.showRelativeTimes ? TimeAgoComp : isoDate} interactive>
```
RomanLamsal1337, line 934: the `interactive` prop enables the selection of the tooltip's value. This ensures that the tooltip's content can be copied.

RomanLamsal1337:
We decided to not add the `showRelativeTimes` to the URL queryparams as the settings only affects the view and not the actual data set which will be displayed.

Might be added to localstorage, any thoughts on this?

lhanke-dev:
Hi @tchiotludo, i'll create a UiOptions property for that and add it to the settings page.

Are you nevertheless okay with keeping the toggle in the filter bar of the table?
I'd say it's easier to be found in that location, especially when one's new to AKHQ.

Summary: UiOption backend + UiOption frontend settings page + Toggle in the table header

tchiotludo:
@lhanke-dev : Perfect, I didn't notice the toggle, I will look at this weekend if possible. 
If it's already here, keep it, If it annoyed me, I will remove myself ;)

lhanke-dev:
I just added the requested settings.

During implementation i noticed that in addition to the topic data view, the topic list and tail view also display relative timestamps.

I'll enhance the PR to respect the setting in both of these views too.

lhanke-dev:
@tchiotludo: The `date-time-format` is now used in TopicList, TopicData and Tail view. I think you can have a look at it again. :)

Issue #579, opened 2021-02-02
Produce to topic from existing message


ghost:
It would be nice to be able to produce a new message into a topic from an existing message as a template.

For example, next to the "Share" button add a new button. On click, the "Produce to topic" view is opened with all the input (value, headers, etc.) copied from the existing message.

Then one can just change a few things and click on "Produce" to insert a new message.

tchiotludo:
good one, PR are welcome

AtulKhanduri:
Would be good to provide support for reproducing message (multiple) to the same topic OR move message to a different queue. One of the use cases is to tackle the failed messages.

Let's say somehow the broker didn't get the acknowledgment from the consumer, which may be due to some issue on the consumer side. So, it'll retry until the maximum retry count is reached, then the message will move over to DLT. Now, when the issue on the consumer side gets resolved, then the failed messages needs to reprocess.

Messages can be reprocessed in two ways:

- by sending the message to the original topic from DLT
- by resending the message in the same original topic

I've attached a video demonstrating the thing which I just described.


https://user-images.githubusercontent.com/5286674/125690843-f2072938-ce48-4d17-9e2f-c41e8582ad80.mov

FYI - Above video is from [KaDeck tool](https://www.xeotek.com).

PR #1044, opened 2022-03-17
feat: Produce to topic from existing message

timonback:
Adds copy button to kafka message in TopicData view, which allows to reproduce the message to the same or any other topic of the cluster.

Does implement the initial ask of #579 .
Does not implement the enhancement as suggested by @AtulKhanduri.


tchiotludo:
Works perfectly !! 
Thanks for the nice contributions

PR #1069, opened 2022-04-08
Close consumer when returning null after calculating offset for newes…

neeraj-singh47:
…t sort

Closes consumer when returning null for getOffsetForSortNewest method.

There are chances of too many open files exception if a significant number of partitions are empty

### DIFF
--- a/src/main/java/org/akhq/repositories/RecordRepository.java
+++ b/src/main/java/org/akhq/repositories/RecordRepository.java
@@ -396,6 +396,7 @@ public class RecordRepository extends AbstractRepository {
                 }
 
                 if (last == partition.getFirstOffset() || last < 0) {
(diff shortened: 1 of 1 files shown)


tchiotludo:
Hey,
It seems that the close must be upper now ?  
Like here with a try for resources ?  
https://github.com/neeraj-singh47/akhq/blob/2ce4cbce2bd6a63aa9d400eaf0512ec00a215165/src/main/java/org/akhq/repositories/RecordRepository.java#L253

neeraj-singh47:
try with resources will close the consumer before it's used for polling and outside the parallel stream it'll be a single consumer instance used by multiple threads. The intention is to close consumer only when null is returned while calculating the offset.

PR #1076, opened 2022-04-20
Upload Connector Config

dstrivelli:
Adds button and file input to the Connect Create page so that a user can optionally upload a connector file (from JSON) and have the form data auto populate.

Changes made:
- Added [hidden] file input (outside of form) on ConnectCreate page
- Added "Upload" button (outside of form) with function `handleFileUpload` to click the file input
- Added function for handling file selection `onFileUpload`, which reads/parses a JSON file and sets state based on the contents.
- Removed line that was automatically resetting the subject name.

### DIFF
--- a/client/src/containers/Connect/ConnectCreate/ConnectCreate.jsx
+++ b/client/src/containers/Connect/ConnectCreate/ConnectCreate.jsx
@@ -438,6 +438,25 @@ class ConnectCreate extends Root {
     reader.readAsText(value.target.files[0]);
   };
 
+  handleDownload = value => {
+    const { formData } = this.state;
+    let filename = formData['subject'] + ".json";
+    let filtered = {};
+    for (let d in formData){ if (formData[d] != '' && formData[d] != null) { filtered[d] = formData[d] }}
+    let text = JSON.stringify({
+      "name": formData['subject'],
+      "config": filtered
+    })
+
+    var element = document.createElement('a');
+    element.setAttribute('href', 'data:application/json;charset=utf-8,' + encodeURIComponent(text));
+    element.setAttribute('download', filename);
+    element.style.display = 'none';
+    document.body.appendChild(element);
+    element.click();
+    document.body.removeChild(element);
+  }
+
   render() {
     const { formData, selectedType } = this.state;
     const { history } = this.props;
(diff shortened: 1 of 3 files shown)


tchiotludo:
Hey @dstrivelli, thanks for this one

Just try quickly, overall it works well. 

Just we need more validation to avoid this error: 
image

And after some reflexion, why not a simple txt file like all examples in kafka docs : 
```
name=local-file-sink
connector.class=FileStreamSink
tasks.max=1
file=/tmp/test.sink.txt
topics=connect-test
```
seems to be more easy since copy paste from the docs will works ? 

Tell me what you think

dstrivelli:
I will take another look at the validation of the `subject` field. My inclination was to use the same format/strategy that Confluent Control Center (C3) uses so that the feel/function is the same. In C3, a user can upload/download an associated JSON, which is ultimately POSTed to the connect API as the same JSON object.

For instance, when we are launching connectors without C3 or AKHQ, we use `cURL` and the JSON connector configuration to deploy new connectors.

JSON Example from Confluent Control Center may look like this.
```
{
    "name": "MirrorSourceConnector",
    "config": {
        "connector.class": "org.apache.kafka.connect.mirror.MirrorSourceConnector",
        "emit.checkpoints.interval.seconds": "1",
        "emit.heartbeats.interval.seconds": "1",
        "groups": "consumer-group-.*",
        "replication.factor": "1",
        "source.cluster.alias": "backup",
        "source.cluster.bootstrap.servers": "vip1:9092",
        "sync.topic.acls.enabled": "false",
        "target.cluster.alias": "primary",
        "target.cluster.bootstrap.servers": "vip2:9092",
        "topics": ".*test-topic-.*",
        ...
    }
}
```

tchiotludo:
OK sound reasonable, go for the json. 
If you just can : 
- add some validation of the json format
- move the button upload near to create button (at the bottom bar)
please ?

dstrivelli:
@tchiotludo my most recent commit (https://github.com/tchiotludo/akhq/pull/1076/commits/fa6f77d61225424757e53fe042c466321fc985c0) addresses the following: 
- Moves upload button to footer bar to the left of `Create` button with some margin for spacing.
- Moves footer bar out of conditional logic so that it's visible all the time (so upload can be clicked)
- Adds additional condition to `validate()` to return false if `formData` or `schema` are empty (so create button is disabled by default until `validate()` is `null`)
- Allows name (`subject`) to be excluded from JSON file, defaults to `''` if not present.
- Adds Joi validation routine for items that appear within connector config JSON, but are not present on plugin definitions.

dstrivelli:
Added a download button too so that people don't have to guess on format and can save their partially configured connector settings.

dstrivelli:
Hey @tchiotludo any additional thoughts?

tchiotludo:
@dstrivelli still have issues : 
- there is error at built time, look at github action 
- image still some error in case of wrong format
- image please move the upload button here

FrankMormino:
This will be a huge QoL improvement for some teams out there. Thanks again for this @dstrivelli  <3

tchiotludo:
@dstrivelli closing due to inactivity, reopen if you could work on it

Issue #1094, opened 2022-05-19
API: return 404 when cluster not configured


lostiniceland:
Currently, when trying to access the API with a wrong (not configred cluster-name), an Exception will be thrown and 500 is returned.

http://localhost:8080/akhq/api/BROKEN/topic/foo/data
`"Internal Server Error: Missing required configuration \"bootstrap.servers\" which has no default value."`

The API handling will silently accept the given cluster-name and forward the request to the Kafka client which in this gets does not receive the proper configuration for `bootstrap.servers` and correctly fails.
Instead, the API should check if the given cluster exists (is configured) and only continue when successfull. Otherwise, short-circuit with a 404.

rdrck47:
Hi, I want to contribute my good first issue. Can I start working on this?

tchiotludo:
For sure 😊

PR #1100, opened 2022-05-25
Added validations for the cluster parameter in API

rdrck47:
Hello! Here is a PR that targets issue [#1094](https://github.com/tchiotludo/akhq/issues/1094).
I think the proposed solution that I developed could be better, I am trying to improve the solution implementation without increasing code complexity, in order to avoid introducing new bugs.

* Adds validations in the API for checking that the provided cluster-name parameter is configured


tchiotludo:
superseded by #1104

PR #1104, opened 2022-05-26
Added validations for the cluster parameter in API using exceptions

rdrck47:
Hello! Here is an alternative solution that targets issue #1094.

* Adds validations in the API for checking that the provided cluster-name parameter is configured using exceptions.


tchiotludo:
perfect, just inverted the logic and use a custom exception! 

Thanks :+1:

Issue #1122, opened 2022-06-19
Float/Double fields are not serialized when containing Integer/Long values.


marcellorego:
I have a AVRO schema where a field is defined as an UNION of `double` or `null` types and default value is `null`.

```
{
      "name": "fFinalVolume",
      "type": [
        "null",
        "double"
      ],
      "default": null
}

```
But when I try to produce data into related topic (using correctly the Value schema definition), and data field contains an integer/long value instead of a float/double, serialization fails.

Data sample:
```
[...]
"fFinalVolume": 320,
"fFinalGrossWeight": 0,
"fFinalNetWeight": 320,
[...]
```

Below you can find part of the stack trace.

```
Suppressed: org.apache.avro.UnresolvedUnionException: Not in union ["null","double"]: 320 (field=fFinalVolume)
		at org.apache.avro.generic.GenericDatumWriter.writeField(GenericDatumWriter.java:223)
		at org.apache.avro.generic.GenericDatumWriter.writeRecord(GenericDatumWriter.java:210)
		at org.apache.avro.generic.GenericDatumWriter.writeWithoutConversion(GenericDatumWriter.java:131)
		at org.apache.avro.generic.GenericDatumWriter.write(GenericDatumWriter.java:83)
		at org.apache.avro.generic.GenericDatumWriter.writeWithoutConversion(GenericDatumWriter.java:145)
		at org.apache.avro.generic.GenericDatumWriter.write(GenericDatumWriter.java:83)
		at org.apache.avro.generic.GenericDatumWriter.writeField(GenericDatumWriter.java:221)
		... 34 common frames omitted
		Suppressed: org.apache.avro.UnresolvedUnionException: Not in union ["null","double"]: 320
			at org.apache.avro.generic.GenericData.resolveUnion(GenericData.java:896)
			at org.apache.avro.generic.GenericDatumWriter.resolveUnion(GenericDatumWriter.java:272)
			at org.apache.avro.generic.GenericDatumWriter.writeWithoutConversion(GenericDatumWriter.java:143)
			at org.apache.avro.generic.GenericDatumWriter.write(GenericDatumWriter.java:83)
			at org.apache.avro.generic.GenericDatumWriter.writeField(GenericDatumWriter.java:221)
			... 40 common frames omitted
```

marcellorego:
https://github.com/tchiotludo/akhq/pull/1123

PR #1123, opened 2022-06-19
Allow integer/long to be serialized as float/double when using AVRO schema.

marcellorego:
fixes https://github.com/tchiotludo/akhq/issues/1122


tchiotludo:
Thanks a lot :+1:

PR #1136, opened 2022-07-03
feat(docker): Add healthcheck

ThomasSanson:


Dockerfile
```diff
@@ -7,6 +7,9 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/* && \
     apt-get clean
 
+HEALTHCHECK --interval=1m --timeout=30s --retries=3 \
+  CMD curl --fail http://localhost/8080 || exit 1
```
tchiotludo, line 11:
```suggestion
  CMD curl --fail http://localhost:8080/health || exit 1
```

tchiotludo (changes requested):
Seems that there is a mistake on the url

Dockerfile
ThomasSanson: indeed 😅

sandropace:
healthchecks are behind authentication.  that should be outside of auth, so that k8s can still verify health of the container running akhq.

tchiotludo:
No, healthcheck are on separated port with no auth

Issue #1101, opened 2022-05-25
Consumer group with slash crashed the ui


baganokodo:
Hi,

I suspect this issue is caused by special chars like "/" and ":" in our consumer group names. Could you confirm that special chars are not accepted in consumer group names?

sample consumer groups, "data/data-serving:forwarder", "local_test/hostname.local"

Thanks

tchiotludo:
can you send a screenshot please?

baganokodo:
Thanks @tchiotludo for the quick response. I feel it quickly popped a red error banner, and the entire page was gone instantaneously. I checked the AKHQ log, nothing special, no errors. The web page is totally blank, it is black mostly like because the chrome theme I selected. Page source only shows some Javascript references. 

Thanks

tchiotludo:
It's seems that there is an hard JavaScript error, can you provide the error from JavaScript console from chrome ?

baganokodo:
![Screen Shot 2022-05-26 at 1 49 11 PM](https://user-images.githubusercontent.com/78040512/170577718-2d43780a-e221-4a94-9513-420d23027180.png)
thanks a lot

baganokodo:
found a good sample on URL encoder, https://www.baeldung.com/java-url-encoding-decoding, I will try to make a PR
thanks very much

rdrck47:
Hello! Is anyone working on this? I would like to take this issue!

tchiotludo:
No one is one it, it's assign to you 👍

PR #1143, opened 2022-07-07
Added encoding and decoding to solve the UI crash when slash characters are used in consumer group names

rdrck47:
Hello! Here is a solution that targets issue [#1101 ](https://github.com/tchiotludo/akhq/issues/1101)

- Adds URI encoding and decoding in the UI to solve the UI crash when slash characters are used in consumer group names.


tchiotludo:
THanks a lot ;)

PR #1152, opened 2022-07-14
feat: reduce security vector attack: remove curl; refact run scripts

vutkin:


tooptoop4 (approved):

tchiotludo:
Yes, I can, but after my holidays 😎
IMO, I would prefer to keep curl and update all package to avoid cve

vutkin:
@tchiotludo could you please review/merge a PR?
Does it make sense to create 2 images: -slim prefixed and regular with curl and other vulnerable tools (just joking)?

vutkin:
As well we could switch to UBI image: registry.access.redhat.com/ubi8/openjdk-11:1.13

vutkin:
> could you please review/merge a PR?
> Does it make sense to create 2 images: -slim prefixed and regular with curl and other vulnerable tools (just joking)?

@tchiotludo could you please review?

tooptoop4:
@vutkin after 'les vacances' time 🌞 🍹

tchiotludo:
@vutkin you change too many things in 1 PR that is not related: 
- add a ubi image not used 
- moved of entrypoint & akhq files, why ? just preference matters it's seems, breaking change  
- changed of user on the docker image: Breaking change 
- removed of healthcheck: Breaking change

We can't merged this one, sorry

PR #1171, opened 2022-08-15
fix: Adding namespace to kubectl port-forward Helm Chart NOTES

rogerioefonseca:
Helm Chart Template NOTES output is not informing the needed to pass the namespace while port forwarding.
image

If we do not pass the namespace it will trigger an error:
image


tchiotludo:
thanks :+1:

PR #1184, opened 2022-08-27
encode groupId to allow '&' as part of the name of a consumer group

owidder:
Hi

Love your tool!!!

But I have problems with consumer group names containing ampersands. Which Kafka allows.
This leads to incorrect URIs to call the backend APIs (e.g., to read the offsets)
To fix this, I encoded the groupIds.

What do you think?

Thanks!!!
Oliver

### DIFF
--- a/client/src/utils/endpoints.js
+++ b/client/src/utils/endpoints.js
@@ -238,34 +238,34 @@ export const uriConsumerGroups = (clusterId, search, pageNumber) => {
 };
 
 export const uriConsumerGroup = (clusterId, groupId) => {
-  return `${apiUrl}/${clusterId}/group/${groupId}`;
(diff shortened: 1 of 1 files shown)


tchiotludo:
hey @owidder, there is already this one: https://github.com/tchiotludo/akhq/pull/1143. 
Maybe we have a conflict? Do you have made some tests?

owidder:
Hi @tchiotludo 
[#1143](https://github.com/tchiotludo/akhq/pull/1143) is sort of the same. But it does not fix the problem for all APIs (e.g., the offset-API, which I ran into).
I tried to fix it directly in endpoint.js to ensure that groupId is always encoded.

Thanks!!!

Oliver

P.S. Sorry did not make tests

tchiotludo:
thanks @owidder

Issue #1199, opened 2022-09-07
Version 0.22.0 Docker image broken, "insufficient memory for the Java Runtime Environment"


erkexzcx:
Version `0.21.0` is working fine, but it's broken on `0.22.0`.

This is what I am getting when opening `docker logs -f akhq`
```
[root@dockerhost _data]# docker logs -f akhq
[0.004s][warning][os,thread] Failed to start thread "GC Thread#0" - pthread_create failed (EPERM) for attributes: stacksize: 1024k, guardsize: 4k, detached.
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Cannot create worker GC thread. Out of system resources.
# An error report file with more information is saved as:
# /app/hs_err_pid8.log
[0.004s][warning][os,thread] Failed to start thread "GC Thread#0" - pthread_create failed (EPERM) for attributes: stacksize: 1024k, guardsize: 4k, detached.
```

Other notes that might help:
- I am using default JVM OPTS.
- Docker image: https://hub.docker.com/r/tchiotludo/akhq
- 8 Kafka clusters (6 servers per cluster), no authorization (only `bootstrap.servers` field is specified).

jheinitz:
Hello!

We are facing the same issue here. We have three clusters and two of them work fine after upgrading to 0.22.0, but the third one is throwing exactly the same error message. The only difference I can see is that the one where it is not working is a RedHat 8.5 VM while the other two are on RedHat 8.6. 

The strange thing is that the VM where it is not working has 32GB RAM while the other two only have 16GB RAM. And there is more free RAM available on the machine with 32GB where it is not working.

Any idea why this happens?

Kind regards

Jens

tchiotludo:
The container start and the ui is reachable or don't start at all ?

jheinitz:
To me it looks like it doesn't even start the JVM. This is the difference between a working instance and a failing instance:

Failing:
```
root@akhq-abn app]# docker run tchiotludo/akhq:0.22.0
[0.003s][warning][os,thread] Failed to start thread "GC Thread#0" - pthread_create failed (EPERM) for attributes: stacksize: 1024k, guardsize: 4k, detached.
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Cannot create worker GC thread. Out of system resources.
# An error report file with more information is saved as:
# /app/hs_err_pid8.log
[root@akhq-abn app]#
```

Working:
```
[...]

PR #1204, opened 2022-09-12
#1199 Change base image to use focal instead of jammy

jheinitz:
hello,

I have changed the base image to use the `focal` instead of `jammy`. This fixes issue #1199 

Would be great if you could also perform some tests and merge this PR into your repo.

Thanks and kind regards

Jens


tchiotludo:
Not really sure downgrading will be a final solution (just a temporary patch that will failed in long term when we will have to upgrade). 

Can you simply start `docker run -it eclipse-temurin:11-jre bash` and try `java -version`, if it's not work, the issues is on downstream and you can report the issue   [here](https://github.com/adoptium/containers/issues)?

jheinitz:
This is what I get:

```
[root@akhq-abn ~]# docker run -it eclipse-temurin:11-jre bash
Unable to find image 'eclipse-temurin:11-jre' locally
11-jre: Pulling from library/eclipse-temurin
2b55860d4c66: Already exists
2ca45fc4c4ca: Already exists
522d9aa14d93: Already exists
7a19e0a42857: Already exists
Digest: sha256:42e4047f19a6582333d5c47e01fbfacb922791dca7c09784c18d550799a8e3b0
Status: Downloaded newer image for eclipse-temurin:11-jre
root@1baeadda9e6b:/# java -version
[0.003s][warning][os,thread] Failed to start thread "GC Thread#0" - pthread_create failed (EPERM) for attributes: stacksize: 1024k, guardsize: 4k, detached.
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Cannot create worker GC thread. Out of system resources.
# An error report file with more information is saved as:
# //hs_err_pid10.log
root@1baeadda9e6b:/#
```

I will open an issue for that and let you know.

Thanks 

Jens

jheinitz:
Hi,

it seems to be a problem with the docker version. I can confirm for RedHat that it works on RedHat 8.6 with a newer docker version than on RedHat 8.5. Please see [this ](https://github.com/adoptium/containers/issues/215#issuecomment-1142046045) comment for details.

Cheers

Jens

Issue #473, opened 2020-10-22
Search / Pagination in topic data


ankon:
It seems when searching in a topic there's a bunch of oddnesses:
1. The pagination control is very misleading, and seems to show the number of messages in total, not in the search result
2. The search term gets dropped when going to the second page
3. The search results seem to be un-ordered/out-of-order

ankon:
I might find time later today to look into these issues.

polarising-java:
@ankon  did you started looking at this?

tchiotludo:
@polarising-java I don't think :)

ankon:
I have not yet found the time for that, no. 

As a work-around you can always try to modify the URL directly to define where you want to start looking though.

polarising-java:
> It seems when searching in a topic there's a bunch of oddnesses:
> 
> 1. The pagination control is very misleading, and seems to show the number of messages in total, not in the search result
> 2. The search term gets dropped when going to the second page
> 3. The search results seem to be un-ordered/out-of-order

The topic data search is using an event stream to push the results into frontend. So the concept of pagination in the search makes no sense, we will have only 1 page with the results. Apply the sort ( NEWEST) in this case will be overkill, we have to consume all records, filter and then sort by date.

tchiotludo:
- on previous akhq version, the search stop after XXX elements and and let you click next to have the next result. 
- Agree that the NEWEST is not possible on search, maybe change to OLDEST when doing search will help people to understand this.
- The search result is sorted by kafka throu all partition, we have no control on that.

xakassi:
Hi, guys!
I'm looking at @polarising-java PR (Improve data search), that's great, thanks!

But I still can't understand why the pagination of search results is impossible (if I understand correctly "So the concept of pagination in the search makes no sense")?
I have a topic with 10000 records, search by a condition which should return almost all data, but search shows continuous loading, and then when I stop it after a few minutes it shows me only 50 records, and never shows more. It seems like not expected behavior. Could you, please, describe more detailed - is it technically difficult to add pagination to search results and load all search results and you do not want to do it in this PR, or I just do not understand something and all workes as expected?
Thanks!

tchiotludo:
@xakassi @polarising-java, pagination works on all akhq version with react : 
- there is a notion of [searchEnd and after](https://github.com/tchiotludo/akhq/blob/0.14.1/assets/modules/datas/search-sse.js)
[...]

PR #1209, opened 2022-09-16
UI - Faster topic-data search & sorting

HailToDodongo:
This PR fixes a few UI issues in the topic-data view:
1. searching in large topics no longer freezes the entire tab
2. sorting by "Newest" works again (removed in #473)
3. correctly updating the table-row react state

---
Using the search on topics with many rows (e.g. one of our topics has ~5M entries @ 7GB) completely freezes the UI.
The search itself it really fast, however the HTTP result streams "percent" values with the current progress, this happens around once per ms.
The UI then tries to update in real-time and freezes, ironically also rendering the  "Stop" Button unusable.
With this PR it will only update in 0.5% increments.
While i didn't change the backend, it might be a good idea to check if it's possible to reduce the progress events to avoid unnecessary traffic.

---

Sorting by newest now works by assuming that sorting by oldest works, and inserting the data in the front instead.
The search itself will still scan the data from oldest to newest, but the data view will be sorted correctly.
To make this work, rows now have a custom ID instead of using the index.
This also fixes an unrelated issue where expanded messages will auto-expand messages with the same index when switching to the next page. 

---

The last issue was a small fix in the model update, which modified the state directly, now it's creating copies to do proper state updates.

### DIFF
--- a/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
+++ b/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
@@ -143,6 +143,9 @@ class TopicData extends Root {
   _startEventSource = (changePage) => {
     let { selectedCluster, selectedTopic, nextPage } = this.state;
 
+    let lastPercentVal = 0.0;
+    const percentUpdateDelta = 0.5;
+
     let self = this;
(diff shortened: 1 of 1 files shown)


tchiotludo:
Thanks a lot :)

Issue #1195, opened 2022-09-01
Show schema name next to version in topic browser


lathspell:
The topic list (https://akhq/ui/kafka-dev/topic/foo/data?sort=Oldest&partition=All) currently only shows the schema version like "Value: 341" in the "schema" column. Would it be possible to show the schema name there?
Especially in topics where multiple domain event types are transported, it would help to distinguish them.

tchiotludo:
nice idea, PR are welcome :)

10xtechie:
@lathspell ,
Did you mean `schema name `or the `subject name`, which is set for creating a schema? 
@tchiotludo ,
There is no attribute or property called `name` in `org.akhq.models.Schema.java`. The most similar one to `name` property or attribute  I found is `subject`. Is my understanding correct?

tchiotludo:
yes @10xtechie it's subject here

10xtechie:
@tchiotludo ,
Thanks! You can assign it to me :)

PR #1217, opened 2022-09-23
issue #1195 - added schema name or schema subject to schema version

10xtechie:
@tchiotludo ,

The required changes are made to show schema name with schema version. It now looks like the below screenshot.

ss1

### DIFF
--- a/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
+++ b/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
@@ -517,6 +517,20 @@ class TopicData extends Root {
       })
   }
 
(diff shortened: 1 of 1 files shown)


tchiotludo:
Sorry for the very late response ! 
Globally it work, but I have some issues: 
- image during the loading and remove the id schema version, can you add a proper loader. 
- On a large cluster with many schemas, the endpoint by id is very very slow and will overload the schema registry. 

In fact this endpoint is useless like that and you must provide the subject: 
image
The same id can be used on many subject and the endpoint will redirect you on the first one random. 
May I ask you to add the subject on `redirectId` from `SchemaController` to have a proper response?

PR #1233, opened 2022-10-17
Added fix for incorrect datatype and missing fields in Produce topic page

GnanaJeyam:
Fixed bug for below two issues

1. Added missing fields error if they produce a json message with Schema and without complete fields.
2. Added Incorrect field name fields in the error while incorrect datatype is added while producing a json message with Schema


tchiotludo:
Perfect thanks for the contribution!

Issue #1215, opened 2022-09-22
fix vulnerability in akhq


hari819:
Hello @tchiotludo  ,

the vulnerability scan ran on latest akhq image has these issues ,  we could not upgrade the libraries since they are not bundled as jar files ,  all the class files from the dependencies are added into the akhq.jar instead.

Please do you have any plans to upgrade the dependencies , i think akhq is internally using micronaut framework which has these libraries bundled,

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/KONDAV~1/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/KONDAV~1/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

akhq | PRISMA-2021-0213 | java | medium | com.fasterxml.jackson.core_jackson-databind | 2.13.0 | 7.5 | fixed in 2.14, 2.13.1, 2.12.6
-- | -- | -- | -- | -- | -- | -- | --
akhq | CVE-2020-36518 | java | high | com.fasterxml.jackson.core_jackson-databind | 2.13.0 | 7.5 | fixed in 2.12.6.1, 2.13.2.1
akhq | CVE-2022-24823 | java | moderate | io.netty_netty-codec-http | 4.1.72.Final | 4 | fixed in 4.1.77.Final
akhq | CVE-2022-24823 | java | medium | io.netty_netty-codec | 4.1.72.Final | 5.5 | fixed in 4.1.77
akhq | CVE-2022-25857 | java | high | org.yaml_snakeyaml | 1.29 | 7.5 | fixed in 1.31
akhq | CVE-2022-38751 | java | medium | org.yaml_snakeyaml | 1.29 | 6.5 | fixed in 1.31
akhq | CVE-2022-38749 | java | medium | org.yaml_snakeyaml | 1.29 | 6.5 | fixed in 1.31
[...]

PR #1235, opened 2022-10-18
Added CustomHttpResponseHeadersFilter

AdiWehrli:
With this filter we can configurably define response headers to be added to each HTTP response.


src/main/java/org/akhq/middlewares/CustomHttpResponseHeadersFilter.java
```diff
@@ -0,0 +1,75 @@
...
+import javax.inject.Inject;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Filter("/**")
+public class CustomHttpResponseHeadersFilter implements HttpServerFilter {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CustomHttpResponseHeadersFilter.class);
+    public static final String PROPERTY_CUSTOM_RESPONSE_HEADERS = "akhq.http.headers.custom-response-headers";
```
tchiotludo, line 24:
```suggestion
    public static final Map<String, String> PROPERTY_CUSTOM_RESPONSE_HEADERS = "akhq.server.custom-response-headers";
```
Let users provide the headers as Map directly and change the variable name

tchiotludo (changes requested):
Almost good 

Please : 
- Add some docs [here](https://github.com/tchiotludo/akhq/blob/dev/docs/docs/configuration/akhq.md)
- Add [some examples here](https://github.com/tchiotludo/akhq/blob/dev/application.example.yml)

src/main/java/org/akhq/middlewares/CustomHttpResponseHeadersFilter.java
AdiWehrli: Well that didn't work in my unit tests.

tchiotludo: Maybe you need to create a true configuration for that one like this [one](https://github.com/tchiotludo/akhq/blob/dev/src/main/java/org/akhq/configs/Connection.java#L30)

AdiWehrli: Well, we cannot assign a String to a Map as you suggested: ```java
public static final Map<String, String> PROPERTY_CUSTOM_RESPONSE_HEADERS = "akhq.server.custom-response-headers";
```

Comment (tchiotludo) [reply src/main/java/org/akhq/middlewares/CustomHttpResponseHeadersFilter.java]: Yes I know I want people to fill it like that: 

```yaml 

akhq:
  server:
    custom-response-headers:
      customheaders: values
      customheaders2: values2

```

tchiotludo: good catch :+1: 
```yaml 
akhq:
  server:
    custom-response-headers:
     - key: "custom-header-1"
       value: "values1a values1b"
```

AdiWehrli: Configuration is now in class org.akhq.configs.Server. Example:
```yaml
akhq:
  server:
    customHttpResponseHeaders:
      - name: "Content-Security-Policy"
        value: "default-src 'none'; frame-src 'self'; script-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self'; frame-ancestors 'self'; form-action 'self'; upgrade-insecure-requests"
      - name: "X-Permitted-Cross-Domain-Policies"
        value: "none"
```

AdiWehrli:
What exactly does this error message mean: `Couldn't find any clusters on your configuration file, please ensure that the configuration file is loaded correctly`?

AdiWehrli:
Finally all checks have passed :)

AdiWehrli:
I added another test for the situation if no custom HTTP response headers were configured.

AdiWehrli:
So, I will not change anything at this moment.

tchiotludo:
merged, thanks a lot :+1:

tchiotludo:
It's available on [dev](https://akhq.io/docs/dev.html) version.
The release will be when [this](https://github.com/tchiotludo/akhq/issues/1215) was done (upgrade to latest micronaut version)

tchiotludo:
Still better to ask for PR instead on building on your own house :smiley_cat:

Issue #1219, opened 2022-10-03
Disabling Topic Configurations


tchiotludo:
### Discussed in https://github.com/tchiotludo/akhq/discussions/1218

<div type='discussions-op-text'>

<sup>Originally posted by **guzelcihad** October  3, 2022</sup>
Hi,

As you can see in the below picture, a user which has reader role can edit this configuration. The applications doesn't restrict to edit this html inputs. I've opened the browser developer tools and saw that when a user changes these form element, there is no request goes to application. This is fine but it creates a feeling that a user can change this settings. Any way to prevent that?

![Screen Shot 2022-10-03 at 12 59 48](https://user-images.githubusercontent.com/18442727/193551543-0cc19183-e3c6-4958-b624-40ba3e717cf6.png)
</div>

GnanaJeyam:
@tchiotludo I think we can disable the input text field whenever we have reader role.
Can you assign this issue to me. I can check.
Between, How can I login to reader role. I can't find the login button in the UI.
<img width="1783" alt="Screenshot 2022-10-18 at 1 22 05 PM" src="https://user-images.githubusercontent.com/28224298/196370406-e0ebc073-5579-4c7b-9948-41f52d510cad.png">

tchiotludo:
The simplest way is to set [reader](https://akhq.io/docs/configuration/authentifications/groups.html) as `akhq.security.default-group`

GnanaJeyam:
@tchiotludo 
Raised PR for this bug.
https://github.com/tchiotludo/akhq/pull/1237

GnanaJeyam:
@tchiotludo Sorry to Bumping this up. Any update or feedback for this PR.. ?

PR #1237, opened 2022-10-20
fix(topicconfig): Disabled edit access for topic config screen for reader mode

GnanaJeyam:
Fix for [#1219](https://github.com/tchiotludo/akhq/issues/1219)

Disabled edit access for all the fields in topic config when current role has reader role.


tchiotludo:
thanks a lot :)

Issue #148, opened 2019-11-20
[feature] keep header when scrolling on list


apellegr06:
Is it possible to add a feature to keep visible the header on all ist when scrolling down ?

tchiotludo:
Don't really understand ?
On the same page when you scroll header disappear ?

Sugaroverdose:
This is about table headers which atm are scrollable, so if you have screen which isn't have enough heigh to display - you'll scroll down and will not see table header

tchiotludo:
ok got it :)

apellegr06:
Great, that's it !

PR #1243, opened 2022-10-26
Make the dev build work on apple m1

lucapette:
First of all, thank you very much for this amazing project. I'm a long time user and big fan of the project!

Yesterday I started reading the source code to understand how you approach some problems (I'm working on a side-project with overlapping challenges. For example, I'm interested in how you "count" records in a topic).

At some point reading didn't do it anymore and I tried to run the project locally which lead to this pull-request.

If you look at the diff, it feels there's a variety of unrelated changes bundled up in one pull-request. But they're all "build related", let me explain.

In practice, what happened is that I run into a few issues one after the other before I could successfully run `gradlew build` (which still "fails" because of eslint errors. More on this later) and `docker-compose -f docker-compose-dev.yml up` (as suggested in the dev.md doc).

So here's the timeline of the problems I run into (with their relative fixes):

- I run into this [Execution failed for platform aarch64 · Issue #148 · siouan/frontend-gradle-plugin · GitHub](https://github.com/siouan/frontend-gradle-plugin/issues/148) which I fixed by updating the node version to latest [LTE](https://twitter.com/nodejs/status/1585029291380977665)
- Once I had the "correct" node version, I immediately run into [javascript - `npm install` fails on node-gyp rebuild with `env: python: No such file or directory` on Mac pro m1 - Stack Overflow](https://stackoverflow.com/questions/73368730/npm-install-fails-on-node-gyp-rebuild-with-env-python-no-such-file-or-direc) which I fixed by replacing the deprecated `node-sass` with `sass` (and also moved it to devDependencies)
- I also had to update the react-test-rendered version (its version number goes on par with react) before I could finally run the build.
- And finally I run into this build error: `Namespace tags are not supported by default. React's JSX doesn't support namespace tags. You can set throwIfNamespace: false to bypass this warning.` and solved it by optimizing the images (two birds with one stone?) with [SVGOMG - SVGO’s Missing GUI](https://jakearchibald.github.io/svgomg/).
- I then "tested" the dev worfklow by "fixing" an import (which I left in the pull request because it seems in line with existing conventions). 

While doing this, I also updated the `docs/docs/dev.md` a little (using [markdownlint-cli](https://github.com/igorshubovych/markdownlint-cli) rules as a loose reference).

With these changes, I can run `docker-compose -f docker-compose-dev.yml up` but still have problems with `gradlew build`.

The problem is that there's a number of [eslint errors](https://gist.github.com/lucapette/cef522e217a0bf7133a0f8b387585885) to fix.

I can fix the eslint errors myself. I enjoy these tasks because they make me more familiar with a code base while allowing me to think "big picture" (it's some sort of "programming meditation" for me).

I'm reaching out now with a draft pull-request because it doesn't makes sense to me to invest more time into this unless: 

1. You're OK with the direction of the changes I already made.
2. This pull-request getting a bigger diff.
3. We agree on how to fix the eslint errors (Personally, I'm not sure I'd do the prop-types ones... but that's my bias toward "solving" that by using TypeScript which is a whole different conversation. Happy to have that one as well :))

If we work on this pr and merge it (I really hope so, I've been meaning to give back to this project of a while now), here's what I'd after it:

- I add a github workflow that ensures no eslint problems get merged anymore.
- I would do the same exact things with docs (by using markdownlint, I use it in all my projects and it's nice to have a "format on save" option in vs code or intellij for this too)

Looking forward to hearing your thoughts on this!


tchiotludo:
Thanks for the very detailed description! 
First of all, most of the comment is really relevant! 

Let's dig inside the issues but most of them will required separated PR to avoid the massive one and a proper git history! (and some change are easy to be merged)

- Update the node version: Go for it, no problem (PR 1)
- replace `node-sass` to `sass`: same no problem, I was thinking there is more tunning on react side (PR2) 
- The image: same no problems (PR3) 
- Update the docs: no problems also (PR4) 

The last one! the update of React!
> I can fix the eslint errors myself. I enjoy these tasks

Oh my god, you will rescue me for a long long pain :pray: In fact, I don't have the time to update the react version because of this lint error, ... and to be honest, it begin to be a real mess right now. If you have the time and the desire, I will be really thankful for a complete life, dependabot is crying on me for too many times!

Tell me what is your thinking and thanks!

lucapette:
yes this makes a lot of sense to me. I opened a draft pr with a long description so I could provide context about all the changes (they do seem disconnected otherwise).

I will create a bunch of little pull-requests in preparation of the React update. Once we get there, we can discuss how to best approach the eslint errors (some have) obvious fixes, others don't.

I'm closing this pr as we don't need it anymore :)

Looking forward to getting all of this done!

Issue #1006, opened 2022-01-28
Decimals are not parsed correctly


florianmutter:
If I have a schema with a decimal like this (example from the code):
```
{
    "name": "Dog",
    "namespace": "org.akhq",
    "type": "record",
    "fields" : [
        {"name": "id", "type": "int"},
        {"name": "name", "type": "string"},
        {"name": "weight", "type" : { "type" : "bytes", "logicalType" : "decimal", "precision" : 4, "scale" : 2}}
    ]
}
```

and I create a message with decimal 1.00 like this:

```
{
  "id": 1,
  "name": "Doggy",
  "weight": 1.0
}
```
 The message is displayed in the extended json view like this:
```
{
  "id": 1,
  "name": "Doggy",
  "weight": 1
}
```

This json can then not be used to send the message again because it cannot serialize 1.

While testing I got two different errors:

```
Uncaught (in promise) 
Object { status: 500, message: "Internal Server Error: Cannot serialize value", stacktrace: null }
```

And

```
Uncaught (in promise) 
Object { status: 500, message: "Internal Server Error: Expected bytes. Got VALUE_NUMBER_INT", stacktrace: null }
​
```

We also had the case that it did accept the message but created a big decimal like `251235313.123` from the input `1`  but I cannot reproduce this.

florianmutter:
Hmm now that I tried it again, just to be sure it worked like expected. They I only wonder how we did manage to create this random number when sending the same message that was displayed again to the topic...

fabiofilz:
Hi,

I am facing the same issue with Double fields.

So if I get one of the messages from AKHQ and try to publish it again, it shows the error as below so I need to add ".0" and it will work.

![image](https://user-images.githubusercontent.com/45266582/151573698-6c01cbf3-099a-4363-bf95-851cf8721c4e.png)

IsaacNoga:
Hi, i will like to know if it's possible to get this issue assigned :>

tchiotludo:
done @IsaacNoga enjoy ;)

neoscaler:
Experiencing the same problems. Any progress in fixing it, @IsaacNoga ?

tristanessquare:
Root cause for the described issue seems to be that the value is parsed and then stringified. See here:

https://github.com/tchiotludo/akhq/blob/e57cc02c790ae833f1e2fd4f77b5cbe8def03c96/client/src/containers/Topic/Topic/TopicData/TopicData.jsx#L929-L934

PR #1246, opened 2022-11-01
fix[#1006]: replaced 'json-bigint' parser with 'lossless-json'

tristanessquare:
This replacement allows for pretty-printing the json data, but still keeping the original values especially for floating numbers and bigints.

After this change a message with content:

```
{ "someFloat": 1.00, "bigNumber": 123456789012345678901234567890 }
```

will be displayed without losing any information from the original message. But still be pretty-printed:

```
{
  "someFloat": 1.00, 
  "bigNumber": 123456789012345678901234567890
}
```

Prior to this change the message would have looked like this:

```
{
  "someFloat": 1,
  "bigNumber": 1.2345678901234567890123456789e+29
}
```


tchiotludo:
perfect thanks @tristanessquare

PR #1254, opened 2022-11-07
Fix eslint warnings 

lucapette:
Built on top of #1251 

Opening it as a draft because it's expected to fail until 1) we merge #1251 2) we figure out the best strategy to do this.

I'll be doing a couple of tests to confirm some hypothesis and then come back here and comment on 2)

### DIFF
--- a/client/.prettierignore
+++ b/client/.prettierignore
@@ -1,0 +1,2 @@
+build
+node_modules
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "akhq-front-end",
-  "version": "0.1.0",
+  "version": "0.23.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "akhq-front-end",
-      "version": "0.1.0",
+      "version": "0.23.0",
       "dependencies": {
         "@date-io/moment": "^2.8.0",
         "@material-ui/core": "^4.9.5",
@@ -16,7 +16,7 @@
         "ace-builds": "^1.4.8",
         "axios": "^0.21.1",
         "axios-progress-bar": "^1.2.0",
-        "bootstrap": "^4.5.0",
+        "bootstrap": "^4.6.1",
         "convert-units": "^2.3.4",
         "css-loader": "^3.4.1",
         "font-awesome": "^4.7.0",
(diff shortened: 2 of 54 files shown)


lucapette:
In the last commit I did at first pass at prop types. I think I should be able to finish this in the morning tomorrow 💪

lucapette:
Done 🎉 

I added prop types everywhere they were missing. 

Now that's done, I'd wait for:

- The next release with important security fixes to be live
- We merge #1251 

Once we're there, I will rebase this branch and add one "noisy" prettier commit (it will "fix" more than 70 files..). 

It's the best way to solve all minor problems with eslint since we run prettier that way and it hasn't been run in a while I think.

I took some notes about the code I read while pushing through the prop types. All small things I can do in tiny pull requests (which is my fav way of contributing... I know it didn't show so far :D) once we're done with this.

Looking forward to next steps!

tchiotludo:
release is done and merge of #1251 is also done !! 
Go crazy man ! :laughing:

lucapette:
It's green now 🎉 🎉 🎉 

While fixing the warnings, I learned a few things:

- the eslint fix command was broken (I fixed it here but did not fix much)
- the prettier configuration doesn't work well which means `npx run prettier --write .` right now fails. More about this later.
- moving from node-sass to sass exposed some warnings in third-party libraries. I fixed the boostrap one by upgrading it. Font Awesome has the same problem but the fix is available only for version 6. 


This pull request is still a little noisier than I'd like but to be fair I'm not sure I could break it down further. After all, it's a lot of small fixes that accumulated over time because the CI part wasn't that strict (before we upgrade react scripts).

I took a lot of notes about small code things. I'll be opening small pull-requests as I find the time (and hopefully also contributed a little to the java part. Would love to get some experience with micronaut).

About prettier, I can (and would love) to fix that myself but it only works if you are comfortable with another noise pull request (TIL we [can ignore the commit](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) which is very nice).

I can open a pull request as soon as we merge this one so it's also easier to see how that looks like. I like the "fix all things in one go" strategy because it makes upcoming pull-requests less noisy. But, of course, we'll do it whatever way works best for you!

tchiotludo:
Amazing work !! Tested, it's seems that you don't break the app, even amazing! 
I've added a small fix that must be due to bootstrap update. 

I've merged that, thanks for amazing work here ! 

for the prettier things, go for it, no worries, and no need to remove from git blame, it's the history of the project, and sometimes we need to add a large noise! 

Since we love large refactoring, maybe you can update react and others dependencies to last version :crossed_fingers: (I know I abuse but we are doing that large change without errors, love that :heart: )

lucapette:
@tchiotludo really happy we merged this! I'll do the prettier thing as soon as I can.

It's funny that you mention the react upgrade I was afraid of bringing that up :D since it's again a "big change". But maybe not... I will look into it 💪

tchiotludo:
To be honest, my main competence is definitely not on React, I'm more fluent with vuejs. 
In react, if it work, I'm fine, If I need to go deeper on the react ecosystem, i'm completely lost. 

But right now, I'm struggling with [large update](https://github.com/kestra-io/kestra/pull/808) of vue (that is not fun) :cry:

lucapette:
@tchiotludo I'm exactly the other way around so I'm happy to help. It feels good to give back to a project I have used for years!

Issue #1257, opened 2022-11-08
Record header keys are unique


raph-af:
Kafka headers should support duplicate keys and keep ordering (https://cwiki.apache.org/confluence/display/KAFKA/KIP-82+-+Add+Record+Headers#KIP82AddRecordHeaders-Keyduplicationandordering)

This is not supported in akhq due to the headers type being a Map (cf. [Record.java](https://github.com/tchiotludo/akhq/blob/7e99277c0766d808ebca1e8f21eadc41f06b7a13/src/main/java/org/akhq/models/Record.java#L46))

PR #1258, opened 2022-11-09
#1257 Support duplicate header keys

raph-af:


tchiotludo:
Thanks for the contribution :star_struck:

PR #1261, opened 2022-11-11
Deserialization error when produce with emtpy key or value

alozano3:
Hi, I am facing a deserialization error when producing to a topic with an empty Key or Value. I am consuming with a Spring Kafka app, and debugging I noticed that when producing an empty key or value with AKHQ, the key/value is not actually null, an array of bytes.

In this screenshot you can see that they are not null.
produce_empty_key

And the error that is causing is 
```
java.lang.IllegalStateException: This error handler cannot process 'SerializationException's directly; please consider configuring an 'ErrorHandlingDeserializer' in the value and/or key deserializer
Caused by: java.nio.BufferUnderflowException: null
```

The provided solution is to check if the key/value is empty. If it is, produce sending a null.


tchiotludo:
thanks :+1:

PR #1263, opened 2022-11-14
fix(auth): oidc group can be a string

### LINKED ISSUE #840: Oauth 2, support role as string or array
Hi, whenever we try to login using pingf , if the user on pingf has only one role/group attached in oidc then it comes as a string instead of an array and akhq don't recognize that role/group, any advice on this ?

oidc clame retuns something like this, login fails
{
"client_id": "some_id",
  "lastName": "some_last_name",
  "firstName": "some_name ",
  "roles": "some_role",
}

oidc clame retuns something like this, login succeeds
{
"client_id": "some_id",
  "lastName": "some_last_name",
  "firstName": "some_name ",
  "roles": ["some_role","another_role"],
}

oidc config is like below

      oidc:
        enabled: true
        providers:
          pingf:
            label: "Login with Pingf"
            username-field: id
            groups-field: roles
            groups:
              - name: some_role
                groups:
                  - admin

z06tbond:
When authenticating through OIDC, if the user belongs to only one group, then the field containing the group claim (`groupsField` in akhq) can either be an array (with one item) or a string, depending on the IdP implementation.
There is no standard.

With the current code, a user with only one group can't log in if the IdP returns a string.

This PR handles this particular case.

See for example the same issue on k8s: kubernetes/kubernetes#33290

Fixes #840

### DIFF
--- a/src/main/java/org/akhq/modules/OidcUserDetailsMapper.java
+++ b/src/main/java/org/akhq/modules/OidcUserDetailsMapper.java
@@ -106,7 +106,7 @@ public class OidcUserDetailsMapper extends DefaultOpenIdAuthenticationMapper {
 
     /**
      * Tries to read groups from the configured groups field.
-     * If the configured field cannot be found or isn't some kind of collection, it will return an empty set.
+     * If the configured field cannot be found or isn't some kind of collection or string, it will return an empty set.
      *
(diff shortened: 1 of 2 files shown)


tchiotludo:
Thanks a lot ! :+1:

PR #1271, opened 2022-11-17
Add Doximity to companies using list.

dezka:


tchiotludo:
Thanks a lot @dezka :+1:

PR #1274, opened 2022-11-18
Fix configuration so that prettier runs correctly on all files

lucapette:
As promised in #1254, I run prettier on the whole frontend codebase. I had to do a few things to make it work correctly:

- I fixed the prettier config so that it runs on all files (letting prettier detect the parser did the trick)
- Updated prettier, eslint to their latest versions
- Because of ☝️ I had to update the babel parser (the previous one didn't work anymore) and update the eslintrc config
- I added a few tasks to help run prettier/eslint locally on the whole codebase

As it's configured right now, the build will fail if:

- there are eslint problems
- files aren't formatted with prettier

Which, as I see it, it's how we make sure we don't need another pr like this any time soon.

I will do the react upgrade next (no idea how hard that is though) and, if you're OK with it, I  will also upgrade "most" deps to a more recent version because 1) I may need to anyway 2) some deps upgrade (like material ui) will fix most of the audit problems npm finds


tchiotludo:
Amazing works (and sorry for the delay) !! :rocket: 
Didn't notice any issues, works perfectly!

tchiotludo:
@lucapette : ready for the next one :boom: ? I love your little PR :love_letter:

lucapette:
@tchiotludo nothing to be sorry about! Very happy to see this merged :) As soon as I find a minute I'll do this:

- I'll go over the notes I collected while working on this pr. Hopefully this will produce _tiny_ pull requests (for real!) as I make sense of them
- I'll try to get a basic idea of what needs to be done to upgrade react. No idea what I'll find here but I'll open an issue as soon as I have something concrete to discuss about it

PR #1275, opened 2022-11-21
fix(doc): [Nit] fix the sample of GroovyCustomClaimProvider

LittleWat:
I updated the following change for clarification:
- removed unnecessary comma
- replaced the variable name
- modified to use the builder


LittleWat:
Thanks!

PR #1277, opened 2022-11-23
Get all values from auth header in header auth

ThomasK33:
When setting up the header auth with [Traefik's OIDC Middleware](https://doc.traefik.io/traefik-enterprise/middlewares/oidc/), Micronauts already parsed the group's header into a list before AKHQ could.

Invoking `request.getHeaders().get[...]` only returned the group list's first entry and not the entire list.
Thus, one could not configure access for a user who's a member of multiple groups.

This change should be fully compatible with the group header being raw yet separated string and a pre-parsed list of strings.


tchiotludo:
didn't test but seems to be a valid one! thanks !!!

PR #1279, opened 2022-11-24
feat: upgrade to Java 17

loicmathieu:
This PR updates AKHQ to Java 17.

@tchiotludo an alternative can be to update only the container to 17 and build it for both 11 and 17 so compatibility for 11 is kept in case someone deploy it via it's JAR and not a Docker container.

We discuss the possibility to provides both a container with the JRE and the JDK, I didn't do it because I have no idea how your container is built and deployed. May I add a second Dockerfile ?


tchiotludo:
The best for the JDK will be to add a `build-arg` to the docker image with a default to jre 17 and to provide another github action for pushing the image with jdk with another tag. 

For the compatibility with 11, I think to force upgrade and to provide only a 17 version for now. Codebase will not use features from 17 for now, and will wait to see if people is complaining on that. Anyway, I'm almost people will complain (when I see people creating ticket with akhq version 0.14 :roll_eyes:)

tchiotludo:
@loicmathieu you don't add the jdk image?

loicmathieu:
@tchiotludo no, I don't know how to do it exactly and don't have time now to investiguate.

If you want I can provide a separate PR for it but you need to explain to me how you make the release, as fas as I understand the workflow tags the images based on GIT tags. As there will not be different GIT tags for JRE vs JDK images more tags should be added manually right ?

tchiotludo:
done in #1615

PR #1289, opened 2022-12-02
use port from config to perform health check

masterobiwan:


tchiotludo:
I don't think this one could work on every case, not all the people have the configuration files with that properties (depending on they deploy it). 
I think we could assume a fixed port on this one (like on the helm charts)

masterobiwan:
Fair point. I pushed a commit which provides a possible solution for this. No worries if you prefer to keep the fixed port.

masterobiwan:
Also I could maybe fix the version of yq to use, to make sure the behaviour when the field is not found in yaml never change.

tchiotludo:
@masterobiwan : still want to improve it?

PR #1293, opened 2022-12-05
fix(1265): Added schema validation for nested record types | Jeyam

GnanaJeyam:
Added Kafka Schema Validation for nested schema structure.


tchiotludo:
perfect, thanks a lot ! :rocket:

Issue #990, opened 2022-01-21
Github SSO / OAuth2 support


morokin:
Hi, I'm trying to configure AKHQ to authenticate against Github Enterprise, but without success, Micronaut itself seems supporting this type of authentication https://micronaut-projects.github.io/micronaut-security/latest/guide/#oauth2-configuration.

Is it possible to add some examples on how it can be done in the README/docs section?

Thanks

tchiotludo:
It's not supported for now. 
As micronaut said, we need to have : 
> An implementation of [OauthAuthenticationMapper](https://micronaut-projects.github.io/micronaut-security/latest/api/io/micronaut/security/oauth2/endpoint/token/response/OauthAuthenticationMapper.html)

It's seems that the doc made a full example of the implementation.  

PR are welcome

bhavrana:
Hey @tchiotludo
Does this issue still need to be worked on? If yes, I would like to contribute.

tchiotludo:
You are assigned 😎

bhavrana:
Hey @tchiotludo
Apologies for my inactivity under this assigned task. I had a chance to get to it today but noticed that there is a closed PR :
https://github.com/moremagic/akhq/pull/1 addressing this issue. 
Can you please let me know if this still needs to be worked on?

tchiotludo:
done in https://github.com/tchiotludo/akhq/pull/1319

PR #1319, opened 2022-12-22
Github SSO / OAuth2 support

moremagic:
Added support for Github SSO/OAuth2

see : https://github.com/tchiotludo/akhq/issues/990


tchiotludo:
Thanks a lot, let's people trying it !! :+1:

PR #1320, opened 2022-12-28
[OIDC] Support for the "Groups field or username field which be inside other fields" case

dongockien92:
In some cases, the groups-field or username-field can be inside other fields.

Ex: The "Token Claim Name" setting on Keycloak 20 has default value is "realm_access.roles"
image

The roles property will be inside the realm_access property
image


tchiotludo:
Hey !! 
- Can you update since there is a conflict? 
- Can you add a relevant unit test please? I'm not sure to understand perfectly!

tchiotludo:
perfect for the rebase, just missing a unit test and it's ok for me!

dongockien92:
Hi @tchiotludo, I'm creating UT, I'll inform you when I completed it.

dongockien92:
@tchiotludo I pushed UT, please help to take a look.

PR #1332, opened 2023-01-04
Uninstall unused dep

lucapette:
I started looking into the react upgrade and stumbled upon this.

If you're ok with it, I want to open a pr to add a [nvmrc](https://github.com/nvm-sh/nvm) file to the client dir with the version of node we're using (`18.12.1`) 

It would help us using the same versions of node/npm. For example, I had to manually go back to node 14 to create this pull request with a small and "clean" diff. Using later versions would run into lockfileversion changes (it has already happened in the repo [here](https://github.com/tchiotludo/akhq/commit/71015d454fe796a924c6e2f4457481b7dcb6fd52) for example).

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -13899,39 +13899,6 @@
         "prop-types": "^15.6.2"
       }
     },
-    "reactstrap": {
-      "version": "8.5.1",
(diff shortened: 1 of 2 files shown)


tchiotludo:
yes no problems!

PR #1335, opened 2023-01-05
Import lodash functions explicitily

lucapette:
Before: 
Screenshot 2023-01-05 at 11 02 50

After: 

Screenshot 2023-01-05 at 11 15 37

I had not planned in doing this but it somewhat happened while I was looking into the react upgrade. 

Not a big improvement (before: 69 Kb, after: 27 Kb) but also no reason against it I think? 

I also did some tiny "gardening" in the files I touched. I can't help it sorry 🙃

### DIFF
--- a/client/src/components/Form/Form.jsx
+++ b/client/src/components/Form/Form.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import Joi from 'joi-browser';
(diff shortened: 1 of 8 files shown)


tchiotludo:
It's always a good things to reduce bundle size! Thanks :zipper_mouth_face:

PR #1324, opened 2023-01-02
KsqlDB support

dweber019:
As stated in https://github.com/tchiotludo/akhq/issues/115 it would be nice to have a KsqlDB integration.

If have a first contribution working. I think it should be extended in the future as there are more awesome feature to implement but for now the following features are implemented:

- List streams, tables, queries
- Show KsqlDB server info
- Execution window for DDL
- Simple query window for DML
- Connect to multiple KsqlDB instances and optionally use basic auth
- Basic set of tests incl. a KsqlDBEmbedded implementation

Here some impressions, starting with the lists
akhq-streams
akhq-tables
akhq-queries

The server info
akhq-server-info

The DDL execution window activated with the button "Execute statement"
akhq-ddl

And the DML window activated with the button "Execute query"
akhq-dml
akhq-dml-result

Looking forward to hear your feedback.

PR #1338, opened 2023-01-06
Fix unneeded chain call

lucapette:
While testing #1324 (I have a suggesting but want to be sure it works first), I realised I broke a few things 🤦🤦🤦 when improving build size. So sorry about that.

On the bright side, the code is now simpler too :)


tchiotludo:
this happen to everyone, even the best :muscle:

Issue #1342, opened 2023-01-08
Some Kafka headers are not displayed if type is not String


gschmutz:
Hi

First, thanks a lot for this great tool!

When using the Dead-Letter topic support of Spring Kafka, then a message which is sent to the dead-letter topic includes the original-partition and the original-offset of the message as Kafka Headers. Partition is of type Integer and Offset of type Long and therefore they are not visible when showing the header details. The same is true for the original timestamp. 

![akhq-with-header](https://user-images.githubusercontent.com/3357942/211204900-8d3b440f-d113-46ef-85f5-2c3587b6eeb9.png)

tchiotludo:
Hi @gschmutz, 

Thanks for the report, but this one is really complicated it seems. 
For now, AKHQ don't have any mapping of type and you failed on the same [issues](https://github.com/tchiotludo/akhq/issues/842) than for the body, but for the headers. 
Except manual configuration for the topic, I don't have any clues to handle that. 
If you have an idea, it's welcome

gschmutz:
@tchiotludo I did understand the difficult for both key and value, as their types are provided by the Kafka producer client. 

To be honest, until now (after checking the javadoc) I realized that the header value is a byte[] and not an Object. Which makes sense, so that it is Interoperable. That of course makes it impossible to handle the value without knowing how it has been serialized. 

First I thought you could display a special value instead of empty but even that will not work because it was only the case because both offset and partition were single digits. I just found it confusing because first I thought the DLQ message did not contain values for these two headers.

Lessons learnt I guess is, that if using headers of type other than String makes it really hard for a consumer to handle in a generic way. Now I understand why all the visualization tools I have tried have the same problem with headers.

gschmutz:
Assuming that the Strings are encoded iin UTF-8, I found this java code here to detect it: <https://stackoverflow.com/questions/1193200/how-can-i-check-whether-a-byte-array-contains-a-unicode-string-in-java>.

```java
    static boolean looksLikeUTF8(byte[] utf8) throws UnsupportedEncodingException
    {
        Pattern p = Pattern.compile("\\A(\n" +
                "  [\\x09\\x0A\\x0D\\x20-\\x7E]             # ASCII\\n" +
                "| [\\xC2-\\xDF][\\x80-\\xBF]               # non-overlong 2-byte\n" +
                "|  \\xE0[\\xA0-\\xBF][\\x80-\\xBF]         # excluding overlongs\n" +
[...]

PR #1352, opened 2023-01-22
Implement handling of Kafka header values of type int, long and short

gschmutz:
This pull request handles the problem raised with #1342.

It enhances the reading of Kafka Header values to treat not only String but also Integer, Short and Long values.


tchiotludo:
THanks a lot :+1:

Issue #907, opened 2021-11-14
Need to Login twice 


shnplr:
Hi Testing Helm deploy with basic auth 

After clicking "Logout" get displayed login page (good). But when logging in again (same or different user) the application show briefly and then goes back to the login screen.  Login 2nd time it is OK.  Why would need to logon twice after logging out?

```
appVersion: "0.19.0"

configuration:
  micronaut:
    security:
      enabled: true
      token:
        jwt:
          signatures:
            secret:
              generator:
                secret: bc45ffbcf429017c46a2724b64907261f46b413e541e7963e6520a19d5d460f8
...
  akhq:
...
    security:
      default-group: no-roles # Default groups for all the user even unlogged user
      # Groups definition
      groups:
        admin: # unique key
          name: admin # Group name
          roles:  # roles for the group
            - topic/read
            - etc
...
      basic-auth:
        - username: admin
          password: 8c6976e5b5410415bde908bd4dee15dfb167a9c873fc4bb8a81f6f2ab448a918 # Password in sha256
          groups: # Groups for the user
            - admin
```

shnplr:
Also, when login page is shown unexpectedly after first login, the dialog not centered (shows toward the right) - its like nav space is used even though the nav is not visible.  Login page is centered correctly when click logout.

twobeeb:
Same here on `dev` branch.

The JWT cookie is present (and correct) after the first authentication if it's anything useful.

tchiotludo:
thanks @loicgreffier for digging.
All must be happening here, it the only set on this localstorage : 
https://github.com/tchiotludo/akhq/blob/b360724baa222d0f2c93ff2e375f012fdd61fdc2/client/src/components/Root/Root.jsx#L13

it may be related to `micronaut.server.context-path`? (I see that you used it) 

May be you try to go deeper @loicgreffier ?

kyleli666:
I have the same issue with `micronaut.server.context-path`
I also feel it's related to https://github.com/tchiotludo/akhq/discussions/1297, where the context-path doubled when oidc login succeeds.

PR #1355, opened 2023-01-31
Replace window pathname by component to fix context-path issues on frontend

AlexisSouquiere:
When using Micronaut context-path, the React router will use this context-path as base path for the frontend.
This leads to issues especially with login (see #907) because, for context-path = /linked, the returnTo will be set to /linked/ui/login/ (the if test doesn't work because of this basename) and will try to redirect the user to /linked/linked/ui/login. Because this path is unknown, user will view the login page twice.

Using the component pathname fixes the issue because we won't take the basename into account to set returnTo and let the React router add it

### DIFF
--- a/client/src/components/Root/Root.jsx
+++ b/client/src/components/Root/Root.jsx
@@ -6,9 +6,10 @@ class Root extends Component {
   cancel = axios.CancelToken.source();
 
   componentWillUnmount() {
-    const pathname = window.location.pathname;
+    /* eslint-disable react/prop-types */
+    const pathname = this.props.location?.pathname;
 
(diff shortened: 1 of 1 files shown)


tchiotludo:
are you sure this syntax `this.props.location?.pathname;` works on jsx? 
Not really sure (really not an expert on jsx)

AlexisSouquiere:
Yep. Indeed, it's "standard" JavaScript https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

tchiotludo:
Don't know it was a standard one, thanks for the information and thanks for the pull request

Issue #1298, opened 2022-12-08
next page failed


as93717913:
I click the next page button in the topic tab it will show no data available,
![image](https://user-images.githubusercontent.com/16937091/206402782-25bda432-be58-44d1-a037-a73426ae6c91.png)
but if I enter same url(with the query string), the result will show correctly.
![image](https://user-images.githubusercontent.com/16937091/206402976-2d518853-245a-49aa-8808-bf87fb85ef60.png)

tchiotludo:
Is there an timeout delay that can prevent the loading of message? 
See [here](https://akhq.io/docs/configuration/akhq.html): try increase this `akhq.topic-data.poll-timeout`

PR #1356, opened 2023-01-31
Remove basePath on next page URI to prevent no data available error

AlexisSouquiere:
GET /topic/{topicName}/data returns in the after variable the URI to call to get the next page of data. When using context-path in Micronaut, the context-path is added in the URI.
image

The problem is that we add it another time in the uriTopicData endpoint.
image

This leads to a /linked/linked/api/... call which fails and show the issue mentionned in #1298.
image

If we keep only the after URI as it is and don't add the basePath, pagination works well with context-path


AlexisSouquiere:
Hi @tchiotludo ! Will you have time to check this other PR please ?

tchiotludo:
prefect thanks :rocket:

PR #1371, opened 2023-02-08
Fix newest sort slowness and wrong record number

AlexisSouquiere:
- Added a check to stop polling the topic when we are the end offset. It prevents user from experiencing a latency = akhq.topic-data.poll-timeout parameter the first time you try to see the topic content with NEWEST sort
- Fixed the first offset of the search that was causing inconsistency between akhq.topic-data.size and the number of records displayed (returning one more record by partition than expected) when using pagination (no issues on the first page but starting at page 2 we were receiving one more record by partition)

### DIFF
--- a/src/main/java/org/akhq/repositories/RecordRepository.java
+++ b/src/main/java/org/akhq/repositories/RecordRepository.java
@@ -324,6 +324,12 @@ public class RecordRepository extends AbstractRepository {
                             filterMessageLength(current);
                             list.add(current);
                         }
+
+                        // End of the partition, we can stop here
+                        if (record.offset() == topicPartitionOffset.getEnd()) {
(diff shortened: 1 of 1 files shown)


tchiotludo:
thanks a lot :rocket:

PavelYadrov:
Hello @tchiotludo, could you please tell, approximately, when is the next release?
We have a problem with Newest sorting and waiting for this fix

tchiotludo:
I've this issue that is blocking the release for now: https://github.com/tchiotludo/akhq/issues/1381

mrMigles:
Hi,
@tchiotludo, we're also interesting in AKHQ version with this fix, and as I can see #1381 already closed.
Could you please help us to understand approximately date of release with this fix? 
Thank you!

Issue #1364, opened 2023-02-03
Failing to identifier leader when connecting to Azure EventHubs


justinmchase:
Azure EventHubs has a "kafka projection" mode. It looks a lot like kafka but has some subtle differences I think.

When attempting to connect to eventhubs akhq has an error:

> 2023-02-03 22:53:57,298 [1 31mERROR[0 39m [35mr-thread-5[0 39m [36mi.m.h.server.RouteExecutor[0 39m Unexpected error occurred: Error encoding object [org.akhq.utils.ResultPagedList@691ab908] to JSON: Leader not found (through reference chain: org.akhq.utils.ResultPagedList["results"]->org.akhq.utils.PagedList[0]->org.akhq.models.Topic["partitions"]->java.util.ArrayList[0]->org.akhq.models.Partition["leader"])

Here is my configuration:

#### application.yaml

```yaml
akhq:
  connections:
    kafka:
      properties:
        bootstrap.servers: "recovery-{{ .Values.global.envName }}.servicebus.windows.net:9093"
        security.protocol: SASL_SSL
        sasl.mechanism: PLAIN
        sasl.jaas.config: org.apache.kafka.common.security.plain.PlainLoginModule required username="$ConnectionString" password="{{ .Values.global.AZURE_EVENT_HUBS_ENDPOINT }}";
```

#### https://github.com/tchiotludo/akhq/blob/dev/src/main/java/org/akhq/models/Partition.java#L40-L46

I believe this is the problem code, I belive that EventHubs is reporting that there are no replicas therefore the code to identify the leader is failing.

```java
    public Node.Partition getLeader() {
        return nodes
            .stream()
            .filter(Node.Partition::isLeader)
            .findFirst()
            .orElseThrow(() -> new NoSuchElementException("Leader not found"));
    }
```

Since there are no replicas, the `.findFirst()` finds nothing. I'm assuming the fix will be to either add the partition itself to the nodes list in the case where there are no replicas, or to return `this.id` in the `getLeader()` function instead of throwing an exception.

tchiotludo:
Event hub support can be cool, but I think it will break on other edge case, PR are welcome, closing due to duplicate of #657

justinmchase:
Can you elaborate on what other edge cases you're worried about? It seems like the fix to make it work with event hubs would be pretty simple and I can test it against local single-instance and also GCP cluster. But if you were to elaborate on edge cases I could probably try to take that into consideration.

[...]

Issue #657, opened 2021-04-07
Problem with sasl.jaas.config value for Event Hub Azure


cactuschibre:
Hello!

I tried to use AKHQ to connect to a Event Hub (Azure) and it does not work with AKHQ. Example of configuration:

```
connections:
  my-cluster:
    properties:
      bootstrap.servers: "DOMAIN.servicebus.windows.net:9093"
      security.protocol: SASL_SSL
      sasl.mechanism: PLAIN
      sasl.jaas.config: org.apache.kafka.common.security.plain.PlainLoginModule required username="$ConnectionString" password="Endpoint=sb://DOMAIN.servicebus.windows.net/;SharedAccessKeyName=KEYNAME;KEYNAME=VEYVALUE";
```
(Example from: https://docs.microsoft.com/fr-fr/azure/event-hubs/event-hubs-quickstart-kafka-enabled-event-hubs)

On the AKHQ launched, I get this error in console:

```
2021-04-07 10:48:20,823 WARN  r-thread-2 .a.k.c.a.AdminClientConfig The configuration 'sasl.jaas.config' was supplied but isn't a known config.
2021-04-07 10:48:22,397 WARN  r-thread-2 org.akhq.log.access        [Date: 2021-04-07T10:48:20.484018+02:00] [Duration: 1913 ms] [Url: GET /api/my-cluster/topic] [Status: 500] [Ip: /127.0.0.1] [User: -]
2021-04-07 10:48:22,414 ERROR Group-1-11 o.a.c.ErrorController      Error for List Log dir
(...)
Caused by: org.apache.kafka.common.errors.UnsupportedVersionException: The broker does not support DESCRIBE_LOG_DIRS
```

With the same configuration in another tool like Conduktor, it works.

tchiotludo:
Hi @cactuschibre, 

I know that azure event hub is not working for now. 
We need to disable (silently) some call this is not compatible with event hub.
Unfortunately, I don't have any azure account in order to test. 

PR are welcome here for someone that have an account :smile:

erkexzcx:
+1

PR #1388, opened 2023-02-24
fix(models): Use the partition leader from partition info

justinmchase:
This doesn't necessarily enable EventHubs completely, but this is attempting to address the first issue hit when attempting to connect to Azure Event Hubs which causes it to crash at startup.

The issue is that in `EventHubs` the replica set is empty and therefore there is no leader in the `this.nodes` list, which blows up when `getLeader()` is called.

This change will find the leader in the constructor instead of in the call to `getLeader()`, if there are no replicas (or replica leaders) it will call `partitionInfo.getLeader()` instead, which hopefully does not return null and hopefully will have the correct logic to identify the partition leader, but if it does then can easily be null checked in `getLogDirSize()` which will safely default to 0 in that case.

## Related To

- https://github.com/tchiotludo/akhq/issues/1364
- https://github.com/tchiotludo/akhq/issues/657

### DIFF
--- a/src/main/java/org/akhq/models/Partition.java
+++ b/src/main/java/org/akhq/models/Partition.java
@@ -29,19 +29,27 @@ public class Partition {
         this.lastOffset = offsets.getLastOffset();
         this.nodes = new ArrayList<>();
         for (org.apache.kafka.common.Node replica : partitionInfo.replicas()) {
-            nodes.add(new Node.Partition(
+            Node.Partition partition = new Node.Partition(
                 replica,
                 partitionInfo.leader().id() == replica.id(),
                 partitionInfo.isr().stream().anyMatch(node -> node.id() == replica.id())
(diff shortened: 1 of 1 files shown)


src/main/java/org/akhq/models/Partition.java
```diff
@@ -27,27 +28,37 @@ public Partition(String topic, TopicPartitionInfo partitionInfo, List<LogDir> lo
...
+            this.nodes.add(partition);
+            if (partition.isLeader()) {
+                this.leader = partition;
+            }
+        }
+        
+        if (this.leader == null)
+        {
+            org.apache.kafka.common.Node leader = partitionInfo.leader();
+            this.leader = new Node.Partition(
```
justinmchase, line 47: Should I `null` check the `leader` here?

PR #1430, opened 2023-03-20
add possibilitiy to search for key/value schema subject

markush81:
Proposal PR: filter records by their key or value schema subject 

In cases where you produce records with different schema subjects into one topic, it would be nice and helpful to filter by subject of the schema.


tchiotludo:
@markush81 can you rebase and fix the check that don't work ?

markush81:
> @markush81 can you rebase and fix the check that don't work ?

@tchiotludo all checks passed.

PR #1432, opened 2023-03-23
Allow to configure (de)serialiser on consumers/producers

gdufrene:
Hello,

I suggest this modification to allow users to be able to override Serializer / Deserializer used by Akhq.
It is usefull when you want different processing per clusterId by providing your compatible Serde (byte[]).
Configure application.yaml and use standard kafka properties on key and value (de)serializer to override ByteArrayDeserializer / ByteArraySerializer.

Providing SerDe seems me the easies way to re-use existing code for users.

Changing KafkaModule ease the way to access  producers and consumers configs.
You can re-use per/cluster and default config logic on your serde.

I tried changing Record class and inject a new interface used to manipulate the String rendering of bytes[].
A lot more code is needed and user needs to implement such a dedicated thing.


src/main/java/org/akhq/modules/KafkaModule.java
```diff
@@ -126,15 +128,7 @@ public AdminClient getAdminClient(String clusterId) throws InvalidClusterExcepti
     }
 
     public KafkaConsumer<byte[], byte[]> getConsumer(String clusterId) throws InvalidClusterException {
-        if (!this.clusterExists(clusterId)) {
-            throw new InvalidClusterException("Invalid cluster '" + clusterId + "'");
```
tchiotludo, line 130: This test must be keep I think

tchiotludo (changes requested):

src/main/java/org/akhq/modules/KafkaModule.java
gdufrene: Hello, any news ?
According my comment, this check seems me redondant,
do you still waiting for this changes ?

PR #1467, opened 2023-04-24
Improve schema loading on produce screen

AlexisSouquiere:
On the produce screen, the 2 dropdown lists for key and value schema are producing unnecessary requests to the schema registry. For each schema the backend is getting the last version (200 schemas = 200 calls). With the PR we will only do 1 call to get the schemas name. 

With a lot of schemas in the registry, the screen will be slow (or even crash if like us you have a API Gateway in front of the schema registry - throttling limit)

Changing the way the dropdown lists are populated by only retrieving the schemas name and lets the backend retrieve only the key/value schema when producing will increase performances on this screen.

Moreover, by default if we have a schema (<topic name>-key or <topic name>-value) we'll select it. I don't know if it was deliberated to not select it by default (but if it is I can remove it)


tchiotludo:
a little failed import it seems `Attempted import error: 'uriAllSchema' is not exported from '../../../utils/endpoints' (imported as 'uriAllSchema').`

AlexisSouquiere:
Sorry it's fixed now

PR #1538, opened 2023-07-19
Fix and enhance search

### LINKED ISSUE #1532: Topic Search Filter not work as expected
Im searching in Kafka Topic over the filter in akhq. 

Search 1: Value // Commercial // Contains 

```
{
  "id": "04e2da4c-26a2-4073-82cb-e90ab334fa0c",
  "model": "Commercial"
}

{
  "id": "fb280b17-76cb-4698-b075-edc9028f70e8",
  "model": "Commercial"
}
```

Search 2: Value // Commercial// Not Contains

```
{
  "id": "04e2da4c-26a2-4073-82cb-e90ab334fa0c",
  "model": "Commercial"
}

{
  "id": "fb280b17-76cb-4698-b075-edc9028f70e8",
  "model": "Commercial"
}
```

shows exact the same records. 

What i expect is that the AKHQ show in Query 2 non of this records.

AKHQ Version: 0.24 
Kafka Version 3.4

AlexisSouquiere:
Fix #1532 

This PR keeps the current behaviour that splits the search string by whitespace to build multiple search patterns. But the way it was supposed to works was confusing. For instance, by giving the following search string `value to search` with Contains option, the result could return records that contain only `to` whereas users were expecting records that contain `value to search`.

- [x] Allow giving a single pattern that contains whitespaces by enclosing it with double quotes (ex: "value to search")
- [x] Support complex patterns like searching for JSON key/value where we need to escape double quotes
- [x] Add details on the way the search works. Especially on the whitespace split + AND/OR operator in case of multiple patterns

On the following example, we are looking for records where the value contains `P1` OR `"field 1":"v1"`

image

### DIFF
--- a/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
+++ b/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
@@ -712,6 +712,14 @@ class TopicData extends Root {
         {this._renderSearchFilter('keySubject', 'Key Subject')}
         {this._renderSearchFilter('valueSubject', 'Value Subject')}
 
+        <p style={{ fontStyle: 'italic' }}>
+          * Whitespaces in search string are considered as separators for search patterns unless
+          enclosed with double quotes
+          <br />
+          In case of multiple patterns, OR operator is applied for Contains / Equals. AND is applied
+          for Not contains
+        </p>
+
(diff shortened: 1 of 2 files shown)


teimyBr:
Seems like the filter now works a way better :D

PR #1539, opened 2023-07-20
Redirect on login page when token expires

### LINKED ISSUE #801: Redirect to login page when user session/token expires
1. I have an AKHQ with security enabled plugged to an LDAP directory. 
2. I successfully log into AKHQ using AKHQ login page
3. I sucessfully use AKHQ ui for a while
4. Something expires: in AKHQ 0.17 I got "500 Internal Server Error" notifications, on AKHQ 0.18 nothing happens anymore but all the screens remain empty. Anonymous connection has no rights at all in my case, so I guess API calls return nothing or timeout.
5. I have to click on the Logout button to show the login page

I'd like the login page to appear automatically when the user token expires, or at least be properly notified that expiration occured and force logout.

AlexisSouquiere:
Fix #801

With the new permissions management, we have now an "Unauthorized" message on the screen when the token is expired
image

I added an interceptor that redirects to the login page

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -1,15 +1,32 @@
 import React from 'react';
 import { BrowserRouter as Router } from 'react-router-dom';
-import { basePath } from './utils/endpoints';
+import { basePath, uriLogin } from './utils/endpoints';
 import Routes from './utils/Routes';
 import { MuiPickersUtilsProvider } from '@material-ui/pickers';
 import MomentUtils from '@date-io/moment';
 import { ToastContainer } from 'react-toastify';
 import { loadProgressBar } from 'axios-progress-bar';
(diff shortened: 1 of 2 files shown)


client/src/App.js
```diff
@@ -1,15 +1,32 @@
...
+        // Used for token expiration by redirecting on login page
+        if (error.response.status === 401) {
+          window.location = uriLogin();
+        }
+
+        return error;
+      } catch (e) {
+        // Propagate cancel
+        if (axios.isCancel(error)) {
+          throw new axios.Cancel('Operation canceled by the user.');
```
alwibrm, line 26: Typo, should be _cancelled_.

PR #1555, opened 2023-08-08
Update Dockerfile

Itiho:
Changes the docker image to use updated alpine in order to solve vulnerabilities present in the ubuntu image.
New image
image

Ubuntu image
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,8 @@
-FROM eclipse-temurin:11-jre
-
+FROM eclipse-temurin:11-jre-alpine
 # install curl
-RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


tchiotludo:
@Itiho this is a large breaking change, you sould provide an alternative image (based on alpine), without changing the default one, and add in the CI/CD the push to this alpine image as a separate tag

tchiotludo (changes requested):
@Itiho this is a large breaking change, you sould provide an alternative image (based on alpine), without changing the default one, and add in the CI/CD the push to this alpine image as a separate tag

tchiotludo:
closing due to inactivity

Issue #1534, opened 2023-07-12
bug: pathType: "ImplementationSpecific" hardcoded 


Syntax3rror404:
Many ingress controllers such as cilium have problems with `pathType: "ImplementationSpecific"`
At the moment we disabled the ingress ressource and made our own with pathType: "Prefix" to make akhq work properly.

I hope you can add a variable for this.

tchiotludo:
for sure, PR are welcome

Issue #1508, opened 2023-06-14
Helm: Ingress with pathType ImplementationSpecific not working


cinimins:
Hi,

the ingress in my cluster is not working. It forwards to `/ui` but then outputs `No server is available to handle this request`.

My values.yaml looks like this (I left out tls and cert-manager/acme annotations):

```yaml
ingress:
  enabled: true
  paths:
    - /
  hotsts:
    - my.host
  annotations:
    ncp/http-redirect: "true"
```

When modifying the ingress to use `pathType: Prefix` (instead of `ImplementationSpecific`), it works. My default ingressclass is nsx, but it also doesn't work with nginx.

My current workaround is to deploy my own ingress instead of using the one from the helm chart. Is there a cleaner way?

libo1987:
Try this

`
{{- if .Values.ingress.enabled -}}
{{- $fullName := include "akhq.fullname" . -}}
{{- $ingressPaths := .Values.ingress.paths -}}
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: {{ $fullName }}
  labels:
    app.kubernetes.io/name: {{ include "akhq.name" . }}
    helm.sh/chart: {{ include "akhq.chart" . }}
    app.kubernetes.io/instance: {{ .Release.Name }}
    app.kubernetes.io/managed-by: {{ .Release.Service }}
  {{- with .Values.ingress.annotations }}
  annotations:
    {{- tpl (toYaml .) $ | nindent 4 }}
  {{- end }}
spec:
  tls:
    - hosts:
      - {{ .Values.ingress.host }}
      secretName: {{ .Values.ingress.secretName }}
  ingressClassName: nginx
  rules:
    - host: {{ .Values.ingress.host }}
      http:
        paths:
        - path: /
          pathType: Prefix
          backend:
            service:
              name: {{ $fullName }}
              port:
                number: 8080
{{- end }}
`

cinimins:
Maybe the ingress controllers in our kubernetes clusters have some weird configuration, I don't know. If it's just a rare corner case, I'm fine with my own ingress, but if not, I was wondering whether it wouldn't be better to make the ingress template a bit more customizable.

The way it is currently implemented with the loop over the [ingressPaths](https://github.com/tchiotludo/akhq/blob/dev/helm/akhq/templates/ingress.yaml#L36)), I don't see an elegant way to make the pathType customizable.

I'm happy to make a PR, I just don't know whether anyone else has the same issue.

[...]

PR #1564, opened 2023-08-24
feat(helm): added ingress pathType

rathberm:
As mentioned in #1534 the ingress `pathType ` was hardcoded to `ImplementationSpecific`, which is not optimal in every environment. I tried to make it possible to change the `pathType` in the values file. What do you think?

I'm using the default template function here as to not break existing setups.

### DIFF
--- a/helm/akhq/templates/ingress.yaml
+++ b/helm/akhq/templates/ingress.yaml
@@ -36,7 +36,7 @@ spec:
 	{{- range $ingressPaths }}
           - path: {{ . }}
             {{- if eq (include "akhq.ingress.apiVersion" $) "networking.k8s.io/v1" }}
-            pathType: "ImplementationSpecific"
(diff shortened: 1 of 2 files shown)


rathberm:
I just saw #1508 where this was discussed too and realized that `$ingressPaths` is a list of Paths which all need their separate `pathType`. As it was already mentioned [here](https://github.com/tchiotludo/akhq/issues/1508#issuecomment-1594835059) I don't see an easy way to solve that. 

With my current solution multiple paths like this:
```
  paths:
    - /
    - /ui
```
will lead to a resource like this:
```
          - path: /
            pathType: ImplementationSpecific
            backend:
              service:
                name: release-name-akhq
                port:
                  name: http

          - path: /ui
            pathType: ImplementationSpecific
            backend:
              service:
                name: release-name-akhq
                port:
                  name: http
```

Every path will always have the same pathType.

tchiotludo:
let's try if it sufficient, since it don't break things. 

THanks for the PR

PR #1582, opened 2023-09-18
CuDo: Update groups.md with warning about recent group changes.

kitor:
Add a warning about recent change, and some pointers where to look if older version is used.

Discussion #1580


tchiotludo:
Thanks for contributions :tada:

PR #1612, opened 2023-11-12
fix(RecordRepository): fix copy infinite loop when record added durin…

### LINKED ISSUE #1377: Copying failing events from DLT to main Topic causes infinite loop
When using the "copy topic" function to copy an event from the DLT to the main topic this causes an infinite loop if the event immediatly fails and is republished to the DLT again.

The "copy topic" function seems to also copy any new events that are published to the topic after the copy action has started. If new events are published to this topic faster than these can be copied this copy action never ends.

This issue can be reproduced reliably by creating two topics with a service consuming events from one topic and publishing the same events to the other topic. Manually add 4 events to the second topic and start the "copy topic" action. Now watch the consumer lag increase rapidly.

valTroadec:
Fix #1377

Hey,

Fix for the infinite loop in the case records are copied to an other topic and an other record has been queued.
I filter them with the last offset we got in the start state.

Can be reproduced easily.

### DIFF
--- a/src/main/java/org/akhq/repositories/RecordRepository.java
+++ b/src/main/java/org/akhq/repositories/RecordRepository.java
@@ -1035,12 +1035,17 @@ public class RecordRepository extends AbstractRepository {
                 );
             }
 
+            Map<Partition, Long> partitionsLastOffsetMap = fromTopic.getPartitions()
+                .stream()
+                .collect(Collectors.toMap(Function.identity(), Partition::getLastOffset));
+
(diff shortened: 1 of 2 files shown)


tchiotludo:
@valTroadec some test are failing, can you fix them please?

tchiotludo (changes requested):
@valTroadec some test are failing, can you fixed them please?

valTroadec:
@tchiotludo I did not see the effects of creating new topics in tests at first. 
I managed to test the copy feature an other way and pushed the test fix.

PR #1650, opened 2024-01-12
Add support for GLUE schema registry (only deserialization)

arindampatra33:
My team at booking.com use Glue as the Schema registry and since AKHQ doesn't  support Glue, we have a hard time visualising avro/json serialised messages in the AKHQ ui.
This PR adds support for deserializing glue . I have tested with json and avro serialized messages and deserialisation worked fine.

Connection configuration looks like this :
     ` schema-registry:
           type: "glue"
           glueSchemaRegistryName: name_of_your_schema_registry
            awsRegion: aws_region_of_your_schema_registry`

Authentication is done using aws default credential provider chain!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -152,6 +152,10 @@ dependencies {
     //AWS MSK IAM Auth
     implementation group: 'software.amazon.msk', name: 'aws-msk-iam-auth', version: '2.0.0'
 
+    // AWS Glue serde
+    implementation ("software.amazon.glue:schema-registry-serde:1.1.15")
+
+
     implementation group: 'io.projectreactor', name: 'reactor-core', version: '3.5.11'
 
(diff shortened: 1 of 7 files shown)


arindampatra33:
@tchiotludo could you look into this PR please

tchiotludo:
@arindampatra33 is it possible to add some unit test (mocking something). 
There is no coverage of the feature for now

tchiotludo:
The best is to add a new classes for that or [here](https://github.com/tchiotludo/akhq/blob/01723b515f42846c01ff8069428121b88ea0cbc7/src/test/java/org/akhq/repositories/AvroWireFormatConverterTest.java).
If you could also add some docs, same as this [one](https://github.com/tchiotludo/akhq/blob/430b80af56528f0d349720b5252c63439d7555e0/docs/docs/configuration/schema-registry/tibco.md)

PR #1670, opened 2024-02-14
sec: Add exact name match on resource

ebrard:
This PR makes it possible to set RBAC on resources based on patterns (as before), on exact name matches and a combinaison of both.


ebrard:
Ok, this can be fully done with regex.

PR #1673, opened 2024-02-17
Enhance support for Glue Schema Registry 

arindampatra33:
Added below functionalities for glue : 
1. show  Key/value schema ID for avro/json/protibuf serialized messages in the topic data page 
2. Disable SchemaRegistry tab if registry is glue  as the functionalities are still not available for glue.
3. Disable on-click on key/value schema IDs in the topic data page for glue.
4. Added doc for configuring glue .

@tchiotludo Could you look into this PR
I have tested by configuring two connections , 1 with Glue , 1 with Confluent 
 I will continue to add more features for glue


tchiotludo:
oups didn't notice, I push a revert :)

Issue #1709, opened 2024-04-05
New Frontend: Spinner blocks results when searching


jonasvoelcker:
Originally my change towards rxjs resulted in a list of entries building up over time. Now there is some loading animation blocking sight:

<img width="1213" alt="image" src="https://github.com/tchiotludo/akhq/assets/47037735/48c184ae-337d-4604-a3a8-1ee55b126317">

FYI: @AlexisSouquiere

PR #1713, opened 2024-04-10
Adding loading animation on topic data search

AlexisSouquiere:
This PR re-add loading animation on topic data search to prevent a "freeze" feeling when applying filters (other than the Search one). When user select a partition, offset or something like this, they will wait few seconds before getting results. But the app never shows the user that the search is in progress. This time it fixes the issues found in #1709 by removing the loading animation the first time we get results in the SSE.

I also added a cancel of pending requests on topic list and data to prevent strange behaviour when user do a search before waiting the previous one to complete + fixed a small navigation issue on topic list (when clicking on Topics menu from the Topic list page)

Finally, I removed the loading check on the search filter that prevent user from using filters when a search is in progress

### DIFF
--- a/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
+++ b/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
@@ -262,7 +262,7 @@ class TopicData extends Root {
 
             if (records.length) {
               const tableMessages = self._handleMessages(records, self.state.sortBy === 'Oldest');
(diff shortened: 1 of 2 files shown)


AlexisSouquiere:
FYI @jonasvoelcker

jonasvoelcker:
@AlexisSouquiere @tchiotludo I just tested it and it works as expected 🙂

Issue #1676, opened 2024-02-23
Empty payload displays null


Chuckame:
# Currently
In the topic view, the kafka events having an empty payload (non null, a byte array of size 0) displays `null`

# Expected
It should display `` (nothing) as it is not a null value (not a tombstone in a compacted topic)

PR #1714, opened 2024-04-13
display blank data instead of null

akshay232:
Resolve Issue:
https://github.com/tchiotludo/akhq/issues/1676

Using nullish coalescing operator to display null only when value is null or undefined, else just display blank.

image


akshay232:
@tchiotludo Could you please take some time to review this PR?

tchiotludo:
good catch, thanks a lot :)

PR #1750, opened 2024-04-16
Fix glue schema registry caching issue

arindampatra33:
The current way of fetching the schema subject name doesn't involve caching and hitting the schema registry multiple times.


arindampatra33:
@tchiotludo  could you merge this when you have time , its a bug fix for glue

PR #1759, opened 2024-04-29
Ability to Add an Authentication Header for External REST API Authentication

cenk1cenk2:
Dear @tchiotludo,

Thank you for your time on still maintaining and improving this project and very excited for the upcoming release.

I have added a draft where it will allow users that are using the external REST authentication method to add a single authentication header while they are doing their requests.

This might become useful, at least in our case is, where the external authentication API is also exposed to the internet, and the user wants to limit this access just through a static token that is passed through the service.

This draft implements this functionality in the most basic form with the ability to add a single header to the request.

I would kindly request you to inform me back and what I can improve if you find this feature also a feasible thing to add to the next release.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,10 +52,12 @@ docker-compose.override.yml
 helm/akhq/README.md
 helm/akhq/LICENSE
 
-
 ## Docs
 docs/node_modules
 docs/.vuepress/.temp
 docs/.vuepress/.cache
 docs/.vuepress/public/contributors.html
 docs/.vuepress/dist
+
+/.factorypath
+/client/.settings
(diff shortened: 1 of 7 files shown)


tchiotludo:
hey @cenk1cenk2, I think it will make more sense to add a generic `headers` that will be a list of custom header using this configuration: 

```yaml
      headers:
        - name: "Authorization"
          value: Bearer your-token
        - name: "X-Api-Key"
          value: "azerty"
```

cenk1cenk2:
Thank you for the guidance @tchiotludo, I have updated it accordingly.

I hope it is up to snuff with my non-existent Java expertise.

Lolo1883:
@tchiotludo 
Hey, we tried the current implementing in prod with those headers and HttpClient and it works!!
if you want to push this forward that would be very great!
Thanks

PR #1773, opened 2024-05-09
Update README.md

jipipayo:
Added CEVA to the list : Who is using AKHQ


jipipayo:
Added CEVA to the list : Who is using AKHQ

Issue #1807, opened 2024-06-04
Regression for handling offset update error


shaneikennedy:
When I try to update an offset for a consumer group (and it fails, as expected) nothing happens but there's an error log in the console. In version 0.24 there used to be an error banner explaining the error

AlexisSouquiere:
Can you explain how you produce this kind of error when updating the offsets ?

shaneikennedy:
For sure

1. Create a topic
2. Create a consumer group for the topic
3. Create a consumer in the consumer group from point 2
4. Produce some data to the topic from point 1
5. Make sure the consumer from point 3 has consumed the data and commited the offsets
6. Try to reset the offset for this consumer group

This should fail because you can't update the offsets of a consumer group while there are consumers connected


But for dev purposes I think you could just find where the request is being made in api.jsx and throw an exception such that the Promise is rejected (in the image you can see that api.jsx:88 is where it happens https://github.com/tchiotludo/akhq/blob/40b4bedb6c047e11205a790a35306211bf78e337/client/src/utils/api.jsx#L88)

shaneikennedy:
Sorry had to remove the image from the issue, i don't think there was anything sensitive but it did have some klarna specifc things in it. Here's a trimmed down image 
<img width="511" alt="image" src="https://github.com/tchiotludo/akhq/assets/24571820/479351ec-6ce9-49c6-a2e5-2610b65b216b">

PR #1809, opened 2024-06-04
fix(ui): broken error handling for all api requests

shaneikennedy:
fixes: https://github.com/tchiotludo/akhq/issues/1807

This is currently affecting all error handling for failed api requests

handleError is not inside a component and so hooks can't be used.

React-router suggests using the redirect function instead of useNavigate but this only works if it's inside a loader/action which I don't think we necessarily are.

Tested and this works, normal error handling shows a toast and if I get a 401/403 im redirected to the login page and i still see a toast with the error message 👍

### DIFF
--- a/client/src/utils/api.jsx
+++ b/client/src/utils/api.jsx
@@ -2,15 +2,12 @@ import React from 'react';
 import axios from 'axios';
 import { toast } from 'react-toastify';
 import 'react-toastify/dist/ReactToastify.css';
-import { useNavigate } from 'react-router-dom';
 
 const configs = {
   withCredentials: true
 };
 
 const handleError = err => {
-  const navigate = useNavigate();
-
   let error = {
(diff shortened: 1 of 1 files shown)


AlexisSouquiere (approved):

AlexisSouquiere:
Well done ! I tested the fix, I confirm that errors are correctly shown now. Thanks 🚀

shaneikennedy:
@AlexisSouquiere Given that this affects every failed api request for everyone on 0.25 do you think we can get a new version for this?

AlexisSouquiere:
I think @tchiotludo agreed in another discussion that we should release more often now. So I assume that after merging this PR and maybe checking few other issues there is in a backlog we will release a patch version

shaneikennedy:
Sounds good, thanks @AlexisSouquiere 🍺

Issue #1448, opened 2023-04-03
tickbox for individual messages in topic to download to csv/json doesn't work


tooptoop4:
i can click the tickbox at the header row to get all messages but clicking individual messages does nothing

dev docker img

PR #1812, opened 2024-06-05
fix(ui): individual message selection for export

gantoin:
This commit makes adjustments in the UI to allow for individual selection of messages for export. As it was requested here: https://github.com/tchiotludo/akhq/issues/1448

demo

### DIFF
--- a/client/src/containers/Topic/Topic/Topic.jsx
+++ b/client/src/containers/Topic/Topic/Topic.jsx
@@ -30,7 +30,8 @@ class Topic extends Root {
     roles: JSON.parse(sessionStorage.getItem('roles')),
     topicInternal: false,
     configs: [],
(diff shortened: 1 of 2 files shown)


client/src/components/Table/Table.jsx
```diff
@@ -90,7 +90,7 @@ class Table extends Component {
                         <>
                           <input
                             type="checkbox"
-                            checked={isChecked}
+                            checked={this.state.exportSome}
```
AlexisSouquiere, line 93: There is no such exportSome state in this component so it's always undefined.
You are still giving the isChecked props in TopicData line 1178 but you removed it here

AlexisSouquiere (changes requested):

AlexisSouquiere (approved):

Issue #1824, opened 2024-06-10
Produce to topic UI allows to select only first 25 topics


jacek-bukowski:
Produce to topic UI allows to select only first 25 topics because it queries only for first page of topics list.

Most likely the return model from backend is not compatible with UI. UI expects `pages` property, while backend returns:
```
{
    "results": [
	...
    ],
    "after": "/api/kfkmqpt/topic?page=2&search&show&uiPageSize=undefined",
    "page": 61,
    "total": 1521,
    "pageSize": 25
}
```

jacek-bukowski:
I've tried the same approach as copy topic UI uses: https://github.com/tchiotludo/akhq/pull/1825

PR #1825, opened 2024-06-10
loading all topics to produce to topic UI #1824

jacek-bukowski:


AlexisSouquiere (approved):
Thanks ! Tested with thousands of topics and it works 🚀

PR #1841, opened 2024-06-19
fix(ui): connect clusters list overflow

AlexisSouquiere:
Reapply style that was removed during the frontend upgrade to fix a small regression on the connect clusters list with a lot of clusters (unable to scroll)

Before:
image

After:
image


AlexisSouquiere:
Fix #1845

PR #1889, opened 2024-08-01
feat(groups): improve consumer groups loading 

### LINKED ISSUE #1863: Unable to display the consumer groups of a topic
Hi,

We used to be able to display the consumer groups existing on a given topic (by clicking on the "Consumer Groups" tab) but it's not working anymore since we migrated our Kafka clusters from Strimzi/OpenShift to Confluent Cloud.

The UI display an "Internal Error" pop-up. 
image

Checking the logs shows that these requests last quite a long time but seem to work since they return a 200. 
image

No idea why there are 2 requests from 2 different threads though.

I tried to set request.timeout.ms to a higher value (1 min then 2 min) but this didn't help.

Any idea what is causing this issue ?

Thanks

### LINKED ISSUE #1919: Failing to display consumer groups if uiPageSize is larger than 94
If uiPageSize>94, the consumer groups information never loads in the topics list.

The browser console logs shows:

TopicList.jsx:296 Uncaught (in promise) TypeError: f.data.filter is not a function
    at TopicList.jsx:296:30
    at Array.forEach (<anonymous>)
    at TopicList.jsx:293:18

Tried v0.24.0 & v0.25.1, same results.

AlexisSouquiere:
- Fix #1863 
- Fix #1919 

Add a dropdown list on the topic screen/consumer groups tab to allow people selecting "Hide empty consumer groups".
This option fixes issues on this screen with cluster containing a lot of consumer groups because we were looping over all the consumer groups and requesting the cluster for their offsets to find the right topics. 

Ignoring empty consumer groups allows AKHQ to give the response quickly because we will search for the topic directly in the member assignation instead of having also the check in the groups offsets (for empty consumer groups)

image

- Add the possibility to choose the default value (like the topic list view) in the application.yml (general settings or by cluster) and in the settings screen

image

- Fix consumer group loading in topic list table
Same issue as the previous one, with a lot of consumer groups, the consumer groups column in the topic list table takes a lot of time to show the result (when you don't reach a timeout).

image

By reusing the new "Consumer Groups Default View", AKHQ will now be able to show the active consumer groups quickly

On a cluster with 1500+ consumer groups, it allows us to finally see (at least) the active groups in few seconds


tchiotludo:
oups, you could rebase and merge @AlexisSouquiere ? 
Thanks a lot :)

AlexisSouquiere:
Ready @tchiotludo ! (finally)

Issue #1916, opened 2024-08-24
Schema name with / character breaks UI


sidtuladhar:
The schema registry below has a schema that has a **/** character in it.

<img width="1217" alt="Screenshot 2024-08-24 at 8 24 18 PM" src="https://github.com/user-attachments/assets/7ca1084b-2c1e-40ad-b9bb-fa5019d2f59b">

After attempting to open `sid/test`, the UI breaks and doesn't load.

<img width="1227" alt="Screenshot 2024-08-24 at 8 24 39 PM" src="https://github.com/user-attachments/assets/46acc653-9124-43b5-b646-c97174aeea3f">

sidtuladhar:
If this issue is relevant, could I be assigned to it? Thanks!

AlexisSouquiere:
Close as completed by #1917

PR #1917, opened 2024-08-27
bugfix: encode schemaId in URL to handle special characters

gantoin:
Related to the issue https://github.com/tchiotludo/akhq/issues/1916

Encoding `schemaId` ensures that special characters in the ID do not break the URL routing.

Capture d’écran 2024-08-27 à 13 01 06


AlexisSouquiere (commented):
Sounds good for merging. Thanks !

PR #1929, opened 2024-09-03
bugfix: update isChecked condition in TopicData table

### LINKED ISSUE #1928: Bug: 'Select All' checkbox doesn't uncheck when a single topic is deselected
When the "Select All" checkbox is selected, all topics are correctly checked. However, if a user manually unchecks a single topic from the list, the "Select All" checkbox should automatically uncheck itself, but it currently remains checked.

demo

gantoin:
Fix #1928

When a user deselects any topic after selecting all, the 'Select All' checkbox will now automatically uncheck itself.

demo2


AlexisSouquiere (approved):
LGTM

Issue #1933, opened 2024-09-05
Documentation: CONSUMER_GROUP without DELETE action


subia:
https://github.com/tchiotludo/akhq/blob/2cab65c5f2a522ac512ff2986abd66f3dc784604/docs/docs/configuration/authentifications/groups.md?plain=1#L51

In the above documentation, under CONSUMER_GROUP resource there is no association with DELETE action, however it is something supported by AKHQ (it enables you the possibility to completely delete an entire consumer group).

The documentation should be updated.

PR #1938, opened 2024-09-12
fix(docs): Update documentation for DELETE action in CONSUMER_GROUP resource

gantoin:
Related to issue: https://github.com/tchiotludo/akhq/issues/1933

This MR updates the documentation to reflect the support for the DELETE action under the `CONSUMER_GROUP` resource in AKHQ. The previous documentation did not mention the ability to delete an entire consumer group, which is a supported feature in AKHQ.

### Changes:
- Added information about the DELETE action in the `CONSUMER_GROUP` section of the documentation.


AlexisSouquiere (approved):
You are right. Even if the consumer group delete feature is not supported in the UI, the endpoint is here

PR #1973, opened 2024-10-11
Fix documentation: connector is missing Update role

### LINKED ISSUE #1974: Fix documentation: connector is missing Update role
# Reason for the change

In the [AKHQ Documentation for Authentication using Groups](https://akhq.io/docs/configuration/authentifications/groups.html), the Connect section in the Roles table does not mention the Update role. This omission may cause confusion, implying that "Connect cannot have this role." However, Connect can indeed have this permission, and without it, updating the connector configuration is not possible.

I propose adding the Update role to the Connect section to clarify this and prevent misunderstandings.

Here is proposed fix: #1973

javidaslan:
Fix #1974 

# Reason for the change

In the [AKHQ Documentation for Authentication using Groups](https://akhq.io/docs/configuration/authentifications/groups.html), the Connect section in the Roles table does not mention the Update role. This omission may cause confusion, implying that "Connect cannot have this role." However, Connect can indeed have this permission, and without it, updating the connector configuration is not possible.

I propose adding the Update role to the Connect section to clarify this and prevent misunderstandings.


AlexisSouquiere (approved):
Good catch, thank you !

PR #1996, opened 2024-11-15
Fixed mistake in role name for Ksqldb aside buttons

Zico56:
Fix a mistake in role name for Ksqldb aside buttons ('KSQDLDB' instead…… of 'KSQLDB').

This is blocking the display of this buttons, and therefore it is not possible to run query or statemetn from web interface.


AlexisSouquiere (approved):
Thanks for catching it

PR #2004, opened 2024-11-20
fix(masking): fixes multi-level nested fields, and hides non-JSON entries

jamfor352:
Hi! Follow up to previous PR as I realised there were a few small issues after deploying to another environment:

1) Returning the original values if it isn't JSON-formattable is actually dangerous, as it means if the schema registry is down or if the wrong schema registry is selected in AKHQ in a multi-schema-registry-setup, the binary format data will still show as a String which in most cases still displays some of the sensitive data. To be safe, this means `json_mask_by_default` will now ONLY show structured data and anything else will display a placeholder message. This prevents any unwanted data leakage.

2) Fixes multi-level nested objects for the 'mask by default' option. Before we weren't prepending `currentKey`. So for example given a request to expose `one.two.three`,  `one.two.three` would still be masked - the filter would have needed to specify just `two.three` (which is incorrect and would also mean `another.two.three` would also be unmasked). This fixes it to work as expected, so `one.two.three` unmasks `one.two.three` and also isn't applied to `another.two.three`

### DIFF
--- a/docs/docs/configuration/akhq.md
+++ b/docs/docs/configuration/akhq.md
@@ -168,6 +168,10 @@ With the above configuration, it will appear as:
 This means, by default, everything is masked.
 This is useful in production scenarios where data must be carefully selected and made available to
 users of AKHQ - usually this is for regulatory compliance of personal/sensitive information.
+
+PLEASE NOTE: This has the side effect of being unable to show unstructured data at all. This is because if the schema registry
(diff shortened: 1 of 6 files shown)


AlexisSouquiere (approved):
Good for me. Thanks for the clarification

jamfor352:
> Good for me. Thanks for the clarification

Thank you!
cc: @tchiotludo I won't be making more changes to this branch so feel free to merge whenever suits :)

PR #2015, opened 2024-12-06
Adding parallel poll and reducing poll timeout

### LINKED ISSUE #1939: List topic messages is extreamly slow after upgrading from 0.25.0 to 0.25.1
Hello

As a follow up of this conversation https://github.com/tchiotludo/akhq/discussions/1880, I'm raising this issue because I couldn't find any issue related to my problem.

Since upgrade to 0.25.1, this API became very slow. The topic I'm listing the content is **empty**.
`2024-09-12 14:27:22,946 INFO io-executor-thread-6 org.akhq.log.access [Date: 2024-09-12T14:25:21.823535418Z] [Duration: 121123 ms] [Url: GET /api/dev/topic/facture.hospi.create/data] [Status: 200]`

Same with 0.25.0 : Still the same empty topic.
`2024-09-12 15:45:15,605 INFO io-executor-thread-13 org.akhq.log.access [Date: 2024-09-12T15:45:14.191436025Z] [Duration: 1415 ms] [Url: GET /api/dev/topic/facture.hospi.create/data] [Status: 200]`

### LINKED ISSUE #1987: akhq 0.25.1 cannot read data from topic
Hi! We faced with the issue when we try to browse topic data. For some kafka clusters I get 503/504 and for some time to getting data increased compare to 0.24.0. No errors in browser's console. No errors when I enable trace logs. Everything what helped us to resolve the issue is rollback to 0.25.0. Can we investigate the root cause?

### LINKED ISSUE #2002: Search messages with sort OLDEST gets super slow if there are any empty topics
Partitions are no longer being polled in parallel when using OLDEST sort (since PR #1843 ). This means that for each search request you have to wait for \<configured-poll-timeout\> * amount of empty topics. For example if the poll-timeout is set to 10 seconds and there are 3 empty partitions then the minimum response time for the request is 30 seconds.

AlexisSouquiere:
Fix #2002 
Fix #1987 
Fix #1939

### DIFF
--- a/src/main/java/org/akhq/repositories/RecordRepository.java
+++ b/src/main/java/org/akhq/repositories/RecordRepository.java
@@ -81,7 +81,7 @@ public class RecordRepository extends AbstractRepository {
     @Inject
     private Masker masker;
 
-    @Value("${akhq.topic-data.poll-timeout:10000}")
+    @Value("${akhq.topic-data.poll-timeout:1000}")
     protected int pollTimeout;
 
     @Value("${akhq.clients-defaults.consumer.properties.max.poll.records:25000}")
@@ -143,7 +143,7 @@ public class RecordRepository extends AbstractRepository {
     private List<Record> consumeOldest(Topic topic, Options options) {
         List<Record> list = new ArrayList<>();
 
-        for (Map.Entry<TopicPartition, Long> partition : getTopicPartitionForSortOldest(topic, options).entrySet()) {
+        getTopicPartitionForSortOldest(topic, options).entrySet().parallelStream().forEach(partition -> {
             Properties properties = new Properties() {{
                 put(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, options.size);
             }};
@@ -171,7 +171,7 @@ public class RecordRepository extends AbstractRepository {
                     }
                 }
             }
-        }
+        });
 
         return list.stream()
             .sorted(Comparator.comparing(Record::getTimestamp))
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -130,7 +130,7 @@ akhq:
 
   topic-data:
     size: 50
-    poll-timeout: 10000
+    poll-timeout: 1000
     kafka-max-message-length: 1000000
 
   audit:


nicolasdelvat:
I tested in local your code with a big enough timeout (20000) to see the long running tasks.
My topic with 5 partitions (4 empty), one with 11 messages, here is the result :
image
Without anything :
image

Looks like there is something done in parallel (not much but its something).

From my understanding if the partition is empty or the offset searched > lastoffset, it waits the timeout.
The last case is reproduced by clicking on this button in the UI ( image )

I filtered out in local those cases (by newest and oldest) and here is the result :
image

I still get messages : 
image

When clicking on the lastoffset button, since there is no new message :
image

How do I procede ? Do I submit a new PR with my changes (without the timeout) and yours ?

In any case, here is the git patch.
[avoid_waiting.patch](https://github.com/user-attachments/files/19094692/avoid_waiting.patch)

ATM, I didn't tests other feature that use the changed function (the copy function use this too).

nicolasdelvat:
Hello,

I did more research, on the parellel stream with our PR (without adding the filters i talked about)
By default the value is the number of CPU allocated to the container/vm -1. So on less than 3 CPU, it will just add overhead of a thread creation without anything. In our dev environment, we probably have a lot of cores, It might not be the case for users.
I'm still using the same topic with 11ich messages and 5 partitions (4 empty).
I added more core to the docker compose (10), still 20000 of timeout.
image
It looks parallel for all threads now
Then using your settings of low timeout:
image
Then I tried 100 as timeout value :
image
It looks like we see diminishing returns from the parallel.

Last test without parallel and 1000 of timeout:
image
So parallel is still better.

As it is, it works but looks like a partial fix for some type of installs :
- for high core count ones (more than 3)
- for users that don't add custom conf (meaning they need to change the conf to see benefits)

@AlexisSouquiere can you affect me this one if you don't have time to test it ?
Or you could merge it and I continue on another PR, as you wish.

AlexisSouquiere:
Thank you for the feedback @nicolasdelvat. I'm marking the PR as ready to be merged. 
Then you can continue on your side the investigation and the improvement on this feature 🚀

Chizhik12:
Hello, have you managed to solve this problem?
Maybe it can be tested somewhere?

PR #2016, opened 2024-12-06
fix(topicdata): rename cluster param to fix data copy authorization check

### LINKED ISSUE #1972: Unauthorized when copying topic in 0.25.1
We updated from 0.24.0 to 0.25.1 and defined the new roles. When I try Copying topic i get 
`Unauthorized: missing permission on resource TOPIC_DATA and action CREATE` and i am logged out. Even though my role definition looks like this:

`admin:
        - resources: ["TOPIC", "TOPIC_DATA", "CONSUMER_GROUP", "CONNECT_CLUSTER"]
          actions: ["READ", "CREATE", "UPDATE", "DELETE", "UPDATE_OFFSET", "DELETE_OFFSET", "READ_CONFIG", "ALTER_CONFIG", "DELETE_VERSION", "UPDATE_STATE", "EXECUTE"]
`

We don't have any patterns or clusters defined in our roles. Am I missing something or is this a bug?

AlexisSouquiere:
Fix #1972 

The SecurityRule wasn't finding any cluster parameter (it was named fromCluster) in the path params. Renaming the parameter fixes the issue

### DIFF
--- a/src/main/java/org/akhq/controllers/TopicController.java
+++ b/src/main/java/org/akhq/controllers/TopicController.java
@@ -605,27 +605,27 @@ public class TopicController extends AbstractController {
 
 
     @AKHQSecured(resource = Role.Resource.TOPIC_DATA, action = Role.Action.CREATE)
-    @Post("api/{fromCluster}/topic/{fromTopicName}/copy/{toCluster}/topic/{toTopicName}")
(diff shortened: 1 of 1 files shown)


drozd0:
Sorry, could you please provide a little bit more details how parameter rename may fix the issue? 
Why you didn't rename `fromTopicName` as well?

PR #2051, opened 2025-01-25
fix(masking): Various performance, functional, and readability fixes for JSON-based masking

jamfor999:
Several fixes here:
1) I realised after further usage and some extensive testing in more complex scenarios that JSON Arrays weren't handled correctly - they were always either masked at the whole field level (on Mask By Default mode), or left unmasked (on Show By Default mode). Now, the same settings that apply now also apply to nested objects inside arrays, or if the array contains primitives, the higher level array will apply. So for example, given:

```
address.firstLine
```

this will now apply to both:

```
{
  "address": {
      "firstLine": "value"
  }
}
```
and now also to all the `firstLine` values in an array type, too (this was not working correctly previously):
```
{
  "address": [
    {
      "firstLine": "value"
    },
    {
      "firstLine": "someOtherValue"
    }
  ]
}
```

Primitives will be handled as expected as well, where a filter like:

```
address.values
```

would now result that the following object:

```
{
    "address": {
      "values": [ "one", "two", "three" ]
    }
}
```

will correctly look like:

```
{
    "address": {
      "values": [ "xxxx", "xxxx", "xxxx" ]
    }
}
```

instead of the current:

```
{
    "address": {
      "values": "xxxx"
    }
}
```
which is currently losing the format, as well as non-sensitive information such as array length.

2) Some readability improvements & fixes, also updating tests to be uniform, so the exact same test code is not repeated. Now the `JsonMaskerTest` interface defines the core tests the input and output, with the `ShowByDefault` and `MaskByDefault` tests implementing the interface (with their own backing `application.yml` which makes them fit the test criteria) and have their own specific tests written in the class only

3) Small performance improvements, including defining the topic -> filtering keys in the constructor rather than constantly re-evaluating
 
4) Remove the phrase "Please contact akhq administrator" which is likely to cause confusion and bring people here, when the issue isn't AKHQ but that a record has not fulfilled the criteria for masking

### DIFF
--- a/src/main/java/org/akhq/utils/JsonMaskByDefaultMasker.java
+++ b/src/main/java/org/akhq/utils/JsonMaskByDefaultMasker.java
@@ -1,8 +1,6 @@
 package org.akhq.utils;
 
-import com.google.gson.JsonElement;
-import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
+import com.google.gson.*;
 import io.micronaut.context.annotation.Requires;
 import jakarta.inject.Singleton;
 import lombok.SneakyThrows;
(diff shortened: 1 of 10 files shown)


jamfor999:
cc: @AlexisSouquiere since you have familiarity reviewing this code beforehand :) A few things changed as explained in the PR. Makes it more reliable and functional and should improve performance (a bit) too

jamfor999:
hi @tchiotludo I was wondering if I could get a quick look on this please? It would be great :)
Thank you

AlexisSouquiere:
@jamfor352 I'm starting to review the PR !

src/main/java/org/akhq/utils/JsonMaskByDefaultMasker.java
```diff
@@ -1,73 +1,123 @@
 package org.akhq.utils;
 
-import com.google.gson.JsonElement;
-import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
+import com.google.gson.*;
```
AlexisSouquiere: Can you avoid using * and import only the specific types ?

src/main/java/org/akhq/utils/JsonShowByDefaultMasker.java
```diff
@@ -1,65 +1,122 @@
...
     private final String jsonMaskReplacement;
+    private static final String ERROR_MESSAGE = "Error masking record";
 
     public JsonShowByDefaultMasker(DataMasking dataMasking) {
-        this.jsonMaskingFilters = dataMasking.getJsonFilters();
         this.jsonMaskReplacement = dataMasking.getJsonMaskReplacement();
+        this.topicToKeysMap = buildTopicKeysMap(dataMasking);
+    }
+
+    private Map<String, List<String>> buildTopicKeysMap(DataMasking dataMasking) {
```
AlexisSouquiere: This method is duplicated in JsonMaskByDefaultMasker

src/main/java/org/akhq/utils/JsonShowByDefaultMasker.java
```diff
@@ -1,65 +1,122 @@
...
-    private final List<JsonMaskingFilter> jsonMaskingFilters;
+    private final Map<String, List<String>> topicToKeysMap;
     private final String jsonMaskReplacement;
+    private static final String ERROR_MESSAGE = "Error masking record";
 
     public JsonShowByDefaultMasker(DataMasking dataMasking) {
-        this.jsonMaskingFilters = dataMasking.getJsonFilters();
         this.jsonMaskReplacement = dataMasking.getJsonMaskReplacement();
+        this.topicToKeysMap = buildTopicKeysMap(dataMasking);
+    }
```
AlexisSouquiere: Can you move all the common variables / constructor upwards in Masker ?

src/main/java/org/akhq/utils/JsonMaskByDefaultMasker.java
```diff
@@ -1,73 +1,123 @@
...
+            logMaskingError(record, e);
+            return createErrorRecord(record);
         }
+    }
+
+    private List<String> getUnmaskedKeysForTopic(String topic) {
+        return topicToKeysMap.getOrDefault(topic.toLowerCase(), Collections.emptyList());
+    }
+
+    private Record createNonJsonRecord(Record record) {
```
AlexisSouquiere: These 3 simple functions have a single reference in maskRecord. I think you can remove them and do the job in maskRecord directly

src/main/java/org/akhq/utils/JsonShowByDefaultMasker.java
```diff
@@ -1,65 +1,122 @@
...
+            return maskJsonRecord(record);
         } catch (Exception e) {
-            LOG.error("Error masking record", e);
+            LOG.error(ERROR_MESSAGE, e);
+            return record;
         }
-        return record;
+    }
+
+    private Record maskJsonRecord(Record record) {
```
AlexisSouquiere: Merge with maskJsonRecord as it's only used here

src/main/java/org/akhq/utils/JsonShowByDefaultMasker.java
```diff
@@ -1,65 +1,122 @@
...
+        record.setValue(root.toString());
         return record;
     }
 
-    private void maskField(JsonObject node, String[] keys, int index) {
-        if (index == keys.length - 1) {
-            if (node.has(keys[index])) {
-                node.addProperty(keys[index], jsonMaskReplacement);
-            }
+    private String[][] preProcessPaths(List<String> maskedKeys) {
```
AlexisSouquiere: Merge with applyMasking

AlexisSouquiere (changes requested):
Can you also add some comments to explain that you handle nested objects aa.bb but that it also applies to arrays ? For maintainability :)

jamfor352:
Hi @AlexisSouquiere , just wanted to leave a comment as I haven't got round to this as I've been taking some time to rest after a period of illness but I'll try my best to get round to these comments pretty soon. Hopefully within the next few days, assuming I'm better.

src/main/java/org/akhq/utils/JsonShowByDefaultMasker.java
jamfor999: > Can you move all the common variables / constructor upwards in Masker ?

Masker is an interface, but I can make an intermediary abstract class, for sure

src/main/java/org/akhq/utils/JsonMaskByDefaultMasker.java
jamfor999: > These 3 simple functions have a single reference in maskRecord. I think you can remove them and do the job in maskRecord directly

while true, I wanted to make the intention behind them more readable as lots of detail in one method can be difficult to parse. But I also appreciate these bits may be overly simple and add too much to the stack, so I'll adjust it

jamfor999:
THanks for the review @AlexisSouquiere , I've addressed your comments :)

jamfor999:
Realised I missed the documentation request. I've added this, and also added more testing as well @AlexisSouquiere , so hopefully it should all be up to scratch now.

jamfor999:
Hi @tchiotludo @AlexisSouquiere is there any chance this could be reviewed soon? My team is itching to get the array issue fixed :)

AlexisSouquiere (approved):
LGTM

jamfor999:
> LGTM

Thank you! 
I think this is ready to merge now @tchiotludo , I won't be updating this PR any more as I'm happy with the state of it

PR #2070, opened 2025-02-10
fix(ui): update date field in produce to topic

mikaelol:
The input field for selecting the timestamp of a record to produce is not updated when clicking the date time picker (but this is a pure UI glitch, because it is the state field `datetime` that is sent to backend regardless of what it says in the input field):

image

### DIFF
--- a/client/src/components/DatePicker/DatePicker.jsx
+++ b/client/src/components/DatePicker/DatePicker.jsx
@@ -102,10 +102,7 @@ class DatePicker extends Component {
 
 DatePicker.propTypes = {
   label: PropTypes.string,
-  value: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.instanceOf(Date)
-  ]),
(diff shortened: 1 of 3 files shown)


client/src/components/DatePicker/DatePicker.jsx
```diff
@@ -102,10 +102,7 @@ class DatePicker extends Component {
 
 DatePicker.propTypes = {
   label: PropTypes.string,
-  value: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.instanceOf(Date)
-  ]),
+  value: PropTypes.oneOfType([PropTypes.string, PropTypes.instanceOf(Date)]),
```
mikaelol, line 105: This (and the comma below) is the result of running `npm run prettier:fix`.

AlexisSouquiere (approved):
LGTM, thanks !
Ok for merging @tchiotludo

PR #2071, opened 2025-02-11
fix(ui): replace moment.js with date-fns

mikaelol:
## What

* Replaced all usage of moment.js with plain javascript Date or date-fns equivalent.
* Installed the official [date-fns/utc](https://github.com/date-fns/utc) to handle UTC not available in core date-fns (which is already installed)
* Renamed client/src/utils/converters.jsx to client/src/utils/converters.js since it does not have any JSX
* Removed the use of `lodash/lowerCase` in converters.js since it is not necessary (built in string lowercase in JS works for this case)
* Created test client/src/utils/converters.test.js to confirm that formatting was same as with Moment.js

## Why

Moment says it quite good at the start of [their readme](https://github.com/moment/moment/blob/develop/README.md):

> Moment.js is a legacy project, now in maintenance mode. In most cases, you should choose a different library.

You already have date-fns installed in this project, and it covers all the usecases you are currently using moment.js for. In addition, Moment.js does not support tree-shaking, so after this dependency is removed some kb is shaved of the final bundle.


client/src/containers/Topic/TopicCopy/TopicCopy.jsx
```diff
@@ -261,23 +260,23 @@ class TopicCopy extends Form {
...
+              minute: timestamp.getMinutes(),
+              second: timestamp.getMinutes(),
+              milli: timestamp.getMilliseconds()
             },
-            'YYYY-MM-DDTHH:mm:ss.SSSZ'
+            "yyyy-MM-dd'T'HH:mm:ss.SSSxxx"
           )
-        : '';
+        : ''; // This case will never happen, 'getTopicOffset' is only called after timestamp is set.
+    // This ternary is probably copy paste from getTopicOffset in ConsumerGroupUpdate.jsx
```
mikaelol, line 279:
```suggestion
        : '';
```
This comment was just added to ask for your opinion. OK to keep the ternary, it won't make a difference with or without.

AlexisSouquiere (approved):
Thanks for removing moment.js !
LGTM @tchiotludo

Issue #2027, opened 2024-12-21
BUG: UI is broken for user with access to all topic metadata but selective access to data


erikgb:
We are running multi-tenant Kafka clusters where we must ensure that tenants cannot read data from other tenants' topics. However, all users should have access to topic metadata in the cluster. Ideally, we want AKHQ to grant access to data in topics based on the ACLs configured for user/groups in the cluster, see https://github.com/tchiotludo/akhq/issues/2025.

Right now, no users have permission to see any topic data through AKHQ. But some of our tenants already have established topic naming standards, so we are trying to set up a POC of access to topic data by applying static patterns in the AKHQ configuration. But this seems broken, and we have also tried using the `dev` image. It's probably related to https://github.com/tchiotludo/akhq/discussions/1910 - as we are seeing the same error message.

Extract of our configuration:

````yaml
default-group: reader-no-data-access
roles:
  reader:
    - resources: ["TOPIC", "TOPIC_DATA", "CONSUMER_GROUP", "CONNECT_CLUSTER", "CONNECTOR", "SCHEMA", "NODE", "ACL", "KSQLDB"]
      actions: ["READ"]
    - resources: ["TOPIC", "NODE"]
      actions: ["READ_CONFIG"]
  reader-no-data-access:
    - resources: ["TOPIC", "CONSUMER_GROUP", "CONNECT_CLUSTER", "CONNECTOR", "SCHEMA", "NODE", "ACL", "KSQLDB"]
      actions: ["READ"]
    - resources: ["TOPIC", "NODE"]
      actions: ["READ_CONFIG"]
groups:
  reader:
    - role: reader
  fifty-reader:
    - role: reader
      patterns: [".*app_fifty_.*"]
  reader-no-data-access:
    - role: reader-no-data-access
````

This is what we observe:
- UI works for an arbitrary user with no extra permissions or relevant groups. User is granted the `reader-no-data-access` role to all topics. ✅ 
- UI works for a user belonging to the reader group. User is granted the `reader` role with access to all topics - including data in topics. ✅ 
- UI is broken for a user belonging to the `fifty-reader` group with the error message "Unauthorized: missing permission on resource TOPIC_DATA and action READ" 💣 

[...]

Issue #1912, opened 2024-08-15
0.25.1 Permissions to limited topics - Unauthorized: missing permission on resource TOPIC_DATA and action READ


thartyhp:
I've since determined this configuration worked okay in 0.25.0, but gets an error in 0.25.1

Behavior in 0.25.0 - User will see all topics as specified in the ACLs, however, if they try to see topic data on a topic they don't have permissions to, the table view will never populate (it may seem like it's loading forever). If it's a topic they have ACL rights to it will display as normal.

Behavior in 0.25.1 - When the user attempts to see the list of topics the table will start to populate, then the "Unauthorized: missing permission on resource TOPIC_DATA and action READ" error is thrown, and the user is sent back to the login page.

### Discussed in https://github.com/tchiotludo/akhq/discussions/1910

<div type='discussions-op-text'>

<sup>Originally posted by **thartyhp** August 14, 2024</sup>
I'm trying to setup a model where we restrict who can see TOPIC_DATA in prod based on group. I can get it so a user can read all TOPIC_DATA or no topic data. But I get an error "Unauthorized: missing permission on resource TOPIC_DATA and action READ" when I try to make it so users can only read select topics based on pattern settings.
 
Me endpoint returns the following. My read of the endpoint payload is that should be what I want.
```
{
    "logged": true,
    "username": “Test User",
    "roles": [
        {
            "resources": [
                "TOPIC"
            ],
            "actions": [
                "READ",
                "READ_CONFIG"
            ],
            "patterns": [
                ".*"
            ],
            "clusters": [
                ".*"
            ]
        },
        {
            "resources": [
                "TOPIC",
                "TOPIC_DATA"
            ],
            "actions": [
                "READ"
            ],
            "patterns": [
                “test-.*”
            ],
            "clusters": [
                ".*"
            ]
        },
        {
            "resources": [
                "TOPIC"
            ],
            "actions": [
                "READ_CONFIG"
            ],
            "patterns": [
                “test-.*”
            ],
            "clusters": [
                ".*"
            ]
        },
        {
            "resources": [
                "NODE"
            ],
            "actions": [
                "READ",
                "READ_CONFIG"
            ],
            "patterns": [
                “test-.*”
            ],
            "clusters": [
                ".*"
            ]
        },
        {
            "resources": [
                "NODE"
            ],
            "actions": [
                "READ",
                "READ_CONFIG"
            ],
            "patterns": [
                ".*"
            ],
            "clusters": [
                ".*"
[...]

PR #2086, opened 2025-02-25
feat: backend send topic permissions and enforce in topic list

mikaelol:
This is a partial solution to https://github.com/tchiotludo/akhq/issues/1912 and https://github.com/tchiotludo/akhq/issues/2027. It’s partial because it is only implemented for the topic list in the frontend. Implementation for the remaining routes can be done, but I wanted to keep this fairly short and hear your opinions on it.

In this MR, every topic fetched from the backend in the topic list API (`api/local/topic`) will contain the permissions that are applicable for that combination of topic, user, and cluster.

<details><summary>See example response (see <code>permissions</code> property of each topic)</summary>

```json
{
    "results": [
        {
            "name": "json",
            "internal": false,
            "partitions": [
                {
                    "leader": {
                        "id": 0,
                        "host": "kafka",
                        "port": 9092,
                        "leader": true,
                        "inSyncReplicas": true
                    },
                    "id": 0,
                    "topic": "json",
                    "nodes": [
                        {
                            "id": 0,
                            "host": "kafka",
                            "port": 9092,
                            "leader": true,
                            "inSyncReplicas": true
                        }
                    ],
                    "logDir": [
                        {
                            "brokerId": 0,
                            "path": "/var/lib/kafka/data",
                            "topic": "json",
                            "partition": 0,
                            "size": 0,
                            "offsetLag": 0,
                            "future": false
                        }
                    ],
                    "firstOffset": 0,
                    "lastOffset": 0,
                    "logDirSize": 0
                }
            ],
            "permissions": {
                "create": false,
                "read": true,
                "update": false,
                "delete": true,
                "readConfig": true,
                "alterConfig": false,
                "topicDataRead": true,
                "topicDataCreate": false,
                "topicDataDelete": true,
                "consumerGroupRead": false
            },
            "logDir": [
                {
                    "brokerId": 0,
                    "path": "/var/lib/kafka/data",
                    "topic": "json",
                    "partition": 0,
                    "size": 0,
                    "offsetLag": 0,
                    "future": false
                }
            ],
            "size": 1,
            "inSyncReplicaCount": 1,
            "streamTopic": false,
            "internalTopic": false,
            "replicaCount": 1,
            "logDirSize": 122
        },
        {
            "name": "test",
            "internal": false,
            "partitions": [
                {
                    "leader": {
                        "id": 0,
                        "host": "kafka",
                        "port": 9092,
                        "leader": true,
                        "inSyncReplicas": true
                    },
                    "id": 0,
                    "topic": "test",
                    "nodes": [
                        {
                            "id": 0,
                            "host": "kafka",
                            "port": 9092,
                            "leader": true,
                            "inSyncReplicas": true
                        }
                    ],
                    "logDir": [
                        {
                            "brokerId": 0,
                            "path": "/var/lib/kafka/data",
                            "topic": "test",
                            "partition": 0,
                            "size": 0,
                            "offsetLag": 0,
                            "future": false
                        }
                    ],
                    "firstOffset": 0,
                    "lastOffset": 0,
                    "logDirSize": 0
                }
            ],
            "permissions": {
                "create": false,
                "read": true,
                "update": false,
                "delete": false,
                "readConfig": true,
                "alterConfig": false,
                "topicDataRead": false,
                "topicDataCreate": false,
                "topicDataDelete": false,
                "consumerGroupRead": true
            },
            "logDir": [
                {
                    "brokerId": 0,
                    "path": "/var/lib/kafka/data",
                    "topic": "test",
                    "partition": 0,
                    "size": 0,
                    "offsetLag": 0,
                    "future": false
                }
            ],
            "size": 1,
            "inSyncReplicaCount": 1,
            "streamTopic": false,
            "internalTopic": false,
            "replicaCount": 1,
            "logDirSize": 122
        }
    ],
    "page": 1,
    "total": 2,
    "pageSize": 25
}
```

</details>

These permissions are used in the TopicList.jsx component to confirm actions for each row, since `roles` object in session storage is the union of all permissions for all topics.

With the permissions in the response listed above, the topic list looks like this with topic "json" and "test":

image

Also fixed a minor bug by only adding `TABLE_DETAILS` when roles TOPIC include READ_CONFIG (I got error in data view when I only had TOPIC_DATA).

### DIFF
--- a/client/src/components/Table/Table.jsx
+++ b/client/src/components/Table/Table.jsx
@@ -151,8 +151,9 @@ class Table extends Component {
 
   renderRow(row, index) {
     const {
-      actions,
+      actions: unconfirmedActions,
       columns,
+      confirmAction,
       extraRow,
       onExpand,
       noRowBackgroundChange,
@@ -161,6 +162,10 @@ class Table extends Component {
       reduce
     } = this.props;
     const { extraExpanded } = this.state;
+    let actions = unconfirmedActions ?? [];
+    if (confirmAction && unconfirmedActions) {
+      actions = unconfirmedActions.filter(action => confirmAction(row, action));
+    }
 
     let extraRowColCollapsed;
     let extraRowColExpanded;
@@ -189,7 +194,6 @@ class Table extends Component {
                 className={column.readOnly ? 'not-allowed' : ''}
                 onDoubleClick={() => {
                   if (
-                    actions &&
                     actions.find(action => action === constants.TABLE_DETAILS) &&
                     !column.expand
                   ) {
@@ -211,7 +215,6 @@ class Table extends Component {
               className={column.readOnly ? 'not-allowed' : ''}
               onDoubleClick={() => {
                 if (
-                  actions &&
                   actions.find(action => action === constants.TABLE_DETAILS) &&
                   !column.expand
                 ) {
@@ -225,7 +228,7 @@ class Table extends Component {
             </td>
           );
         })}
-        {actions && actions.length > 0 && this.renderActions(row)}
+        {unconfirmedActions && unconfirmedActions.length > 0 && this.renderActions(row)}
       </tr>
     ];
     if (
@@ -338,8 +341,12 @@ class Table extends Component {
   }
 
   renderActions(row) {
-    const { actions, onAdd, onDelete, onEdit, onRestart, onShare, onDownload, onCopy, idCol } =
+    const { actions: unconfirmedActions, confirmAction, onAdd, onDelete, onEdit, onRestart, onShare, onDownload, onCopy, idCol } =
       this.props;
+    let actions = unconfirmedActions ?? [];
+    if (confirmAction && unconfirmedActions) {
+      actions = unconfirmedActions.filter(action => confirmAction(row, action));
+    }
 
     let idColVal = idCol ? row[this.props.idCol] : row.id;
 
@@ -357,6 +364,8 @@ class Table extends Component {
               <FontAwesomeIcon icon={faSearch} />
             </span>
           </td>
+        ) || unconfirmedActions.find(el => el === constants.TABLE_ADD) && (
+          <td className="khq-row-action khq-row-action-main" />
         )}
         {actions.find(el => el === constants.TABLE_DETAILS) && (
           <td className="khq-row-action khq-row-action-main action-hover">
@@ -364,6 +373,8 @@ class Table extends Component {
               <FontAwesomeIcon icon={faSearch} />
             </Link>
           </td>
+        ) || unconfirmedActions.find(el => el === constants.TABLE_DETAILS) && (
+          <td className="khq-row-action khq-row-action-main" />
         )}
         {actions.find(el => el === constants.TABLE_CONFIG) && (
           <td className="khq-row-action khq-row-action-main action-hover">
@@ -371,6 +382,8 @@ class Table extends Component {
               <FontAwesomeIcon icon={faGear} />
             </Link>
           </td>
+        ) || unconfirmedActions.find(el => el === constants.CONFIG) && (
+          <td className="khq-row-action khq-row-action-main" />
         )}
         {actions.find(el => el === constants.TABLE_DELETE) && (
           <td className="khq-row-action khq-row-action-main action-hover">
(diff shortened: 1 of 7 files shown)


client/src/components/Table/Table.jsx
```diff
@@ -395,6 +407,8 @@ class Table extends Component {
               <FontAwesomeIcon icon={faGear} />
             </span>
           </td>
+        ) || unconfirmedActions.find(el => el === constants.CONFIG) && (
```
mikaelol:
```suggestion
        ) || unconfirmedActions.find(el => el === constants.TABLE_CONFIG) && (
```

Should correspond to the constant above

src/main/java/org/akhq/models/Topic.java
```diff
@@ -121,6 +122,14 @@ public long getSize(int partition) {
         throw new NoSuchElementException("Partition '" + partition + "' doesn't exist for topic " + this.name);
     }
 
+    public void setPermissions(TopicPermissions permission) {
```
erikgb, line 125: It looks like this class has an immutable design. Instead of providing a setter, could it make sense to add this new info to the constructor? Or provide the topic security model as a "sibling" to the list of topics in the response to the front end? Probably as a map structure with the topic name as the key. I would consider the latter the cleanest solution - if possible to implement.

src/main/java/org/akhq/controllers/TopicController.java
```diff
@@ -103,13 +103,30 @@ public ResultPagedList<Topic> list(
...
+                checkIfClusterAndResourceAllowed(cluster, topic.getName(), Role.Action.READ, Role.Resource.TOPIC),
+                checkIfClusterAndResourceAllowed(cluster, topic.getName(), Role.Action.UPDATE, Role.Resource.TOPIC),
+                checkIfClusterAndResourceAllowed(cluster, topic.getName(), Role.Action.DELETE, Role.Resource.TOPIC),
+                checkIfClusterAndResourceAllowed(cluster, topic.getName(), Role.Action.READ_CONFIG, Role.Resource.TOPIC),
+                checkIfClusterAndResourceAllowed(cluster, topic.getName(), Role.Action.ALTER_CONFIG, Role.Resource.TOPIC),
+                checkIfClusterAndResourceAllowed(cluster, topic.getName(), Role.Action.READ, Role.Resource.TOPIC_DATA),
+                checkIfClusterAndResourceAllowed(cluster, topic.getName(), Role.Action.CREATE, Role.Resource.TOPIC_DATA),
+                checkIfClusterAndResourceAllowed(cluster, topic.getName(), Role.Action.DELETE, Role.Resource.TOPIC_DATA),
+                checkIfClusterAndResourceAllowed(cluster, topic.getName(), Role.Action.READ, Role.Resource.CONSUMER_GROUP)
+            );
```
erikgb: I consider positional parameters of the same type an anti-pattern because it's so easy to make stupid errors also easy to overlook in review. A builder often works a lot better. Is this an option here? Is Lombok available? It can "generate" builders.

mikaelol: I have changed `TopicPermissions` from record to class with lombok builder pattern / annotation.

mikaelol:
@AlexisSouquiere ,  have you had a chance to look at this?

AlexisSouquiere:
Hello @mikaelol. Really sorry for the delay. IMO, I think we should benefit from the /api/me endpoint response that already give the pattern/cluster information. Like this, no updates needed in the backend. Otherwise, it's kind of breaking the single responsibility pattern by not only giving the topics, but also information related to the user's permission.

On your side, I understand it involves updates on several parts of the frontend (consumer groups, schema, connectors, etc.) that you are maybe not familiar with, potential regressions, etc. Maybe we can do this in 2 steps:
- 1st PR: for your part, the topic part, instead of updating the `roles` from the local storage as is, create a `permissions` variable that uses roles + patterns + clusters from the /api/me endpoint. Then update the topic part to use it.
- 2nd PR: update the other screens (consumer groups, connectors, etc.) by switching from the roles to the permissions variable. I can take it

Like this you are sure that you won't cause any regression when implementing it for the topics. Moreover, after having a quick look, I think we should avoid calling the /api/me endpoint for every screen and maybe juste load it after the login.

What do you think ?

mikaelol:
Thanks for the thorough feedback, @AlexisSouquiere, and for your suggested path forward. I see your point about breaking the single responsibility.

I will look into your suggestion for the first PR and see how it works.

Issue #2093, opened 2025-03-04
Slow loading of data when empty partition


nicolasdelvat:
Hello,

I noticed since the poll-timeout got bigger. 
Our dev environment got really slow when fetching data.

The pattern is always the same, more partitions empty = more loading time.
I searched into the source code and it seems the loop on partitions is waiting poll-timeout x partition empty.

Here is an example of a topic with one message and 5 partitions (10s x 4 + fetchmessagetime):
![Image](https://github.com/user-attachments/assets/2fe5ca8a-db77-4ad2-9ddb-493b0b89ddb0)
You can also spot this in the gradle tests, in local some tests are waiting 10s to complete.

I don't think lowering the timeout will resolve anything, but skipping empty partition might do the trick.

@AlexisSouquiere Can you assign this issue to me if you don't see any objection ?
I think, I can fix it without too much effort.

Thank you.

AlexisSouquiere:
Hello, can you have a look to #2015 first ? From what I started to do, lowering to timeout + adding parallel partition poll solves the issue. 

There is no reason to wait the poll timeout if there is no records in the partition. The poll javadoc says _"This method returns immediately if there are records available. Otherwise, it will await the passed timeout. If the timeout expires, an empty record set will be returned"_  So lowering it to like 1s improves the response time. But, depending on the number of partitions, the response time can still be high because partitions are processed sequentially.

I didn't have time to complete all the tests I wanted to do but I'm curious to have your feedback about it. By doing that that we will have like 1s delay before showing empty results

Indeed, your proposal of skipping empty partitions can be the second  step

nicolasdelvat:
You are right, this is a duplicate of #1939.
I will continue the conversation on the PR and test but I think there is a subtlety when partitions are empty, the phrasing looks ambigious in the javadoc.

PR #2100, opened 2025-03-08
fix(topicdata): Avoid waiting the poll timeout when there is no data in the topic partition (#2093 #1939 #2002 #1987)

nicolasdelvat:
Follow up to #2015.

The goal is to avoid waiting the poll timeout configured by the user when the topic/data get messages from empty partitions.

Example of URLs with the issue:
- http://localhost:4000/ui/local/topic/{topic}/data?sort=Oldest&partition=All
- http://localhost:4000/ui/local/topic/{topic}/data?sort=Newest&partition=All

### DIFF
--- a/src/main/java/org/akhq/repositories/RecordRepository.java
+++ b/src/main/java/org/akhq/repositories/RecordRepository.java
@@ -143,7 +143,18 @@ public class RecordRepository extends AbstractRepository {
     private List<Record> consumeOldest(Topic topic, Options options) {
         List<Record> list = new ArrayList<>();
 
-        for (Map.Entry<TopicPartition, Long> partition : getTopicPartitionForSortOldest(topic, options).entrySet()) {
+        getTopicPartitionForSortOldest(topic, options).entrySet().parallelStream()
+            .forEach(partition -> {
+
+            // Skip partition without data
+            if (topic.getPartitions()
+                    .stream()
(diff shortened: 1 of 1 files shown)


AlexisSouquiere:
Still a draft or ready for review @nicolasdelvat ?

nicolasdelvat:
Yep, still in draft.
The tests don't pass, I'm trying things in local.

To explain what I found/trying to do :
On both consumeOldest/Newest, I want to never wait for poll timeout in the following cases
- the partition is empty

The major problem I have is understanding the integration test, I'm not well versed in those.
I need to find if I broke something or they needs ajustements.

There is also those cases that are triggering the timeout: 
- the offsets searched needs to be within firstoffset/lastoffset
- I think this includes the after option

I may do these later, it seems a little hard to me to do everything in this PR.

nicolasdelvat:
I tested with multiple topics :

100 empty partitions :
image

image

12 empty partitions/20 :
image
image

0 empty partition: 
image
image

It looks fine.

On the test side : 
 getAll() seems flaky.
All other tests seems to have a mocks that are not based on reality, so they crash on edge cases and needs to be adjusted.
I will continue to search how to fix them.

Saaharjit007:
Hello, Thanks for working on this issue. Are there any updates on when this PR can be merged? We, too, are unable to upgrade AKHQ since moving to version >0.24 crashes our instance due to timeout errors.

PR #2113, opened 2025-03-11
fix(ui): set pagination to 1 when no data displayed

nicolasdelvat:
Hello,

I noticed, that empty queries show 0 in pagination.   
It's a quick fix since the initial object in js is already set with the good value.
Before :
image
After :
image

Here is the list of urls:
http://localhost:4000/ui/local/connect/connect
http://localhost:4000/ui/local/ksqldb/ksqldb/streams
http://localhost:4000/ui/local/ksqldb/ksqldb/tables
http://localhost:4000/ui/local/ksqldb/ksqldb/queries
http://localhost:4000/ui/local/schema
http://localhost:4000/ui/local/topic <- 2 bugs, one coming from the backend, I fixed it in the js since the impact would be much bigger

### DIFF
--- a/client/src/containers/Connect/ConnectList/ConnectList.jsx
+++ b/client/src/containers/Connect/ConnectList/ConnectList.jsx
@@ -122,7 +122,7 @@ class ConnectList extends Root {
         );
       });
     } else {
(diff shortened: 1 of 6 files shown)


AlexisSouquiere (approved):
LGTM !

Issue #1154, opened 2022-07-20
Format access-log


ericmagna:
Hi,

With the file 
Hello,

In the application.example.yml file I have an example for formatting the accesslog file.

I would like to add the call size.

How can I make this addition?

Thank you for your help.

**Extracting the example configuration**

```yaml
akhq:
  server:
    access-log: # Access log configuration (optional)
      enabled: true # true by default
      name: org.akhq.log.access # Logger name
      format: "[Date: {}] [Duration: {} ms] [Url: {} {}] [Status: {}] [Ip: {}] [User: {}]" # Logger format
```

tchiotludo:
Not for now, PR are welcome

10xtechie:
@ericmagna , @tchiotludo ,
What is `call size` ?
as far as I know, there is no option for call size in `MessageFormatter` of `slf4j`.

tchiotludo:
I don't know what is call size also 😂

ericmagna:
The size is the object returned by the client, excluding headers. Under apache http, the log formatting is %b.

This option allows you to analyze the behavior of calls from customers. For example, customers using the API remain exposed by AKHQ. Some customers may make very important calls.

The access_log file put in elasticsearch makes it possible to finely analyze these calls by customers via a kibana/grafana.

ericmagna:
In the HttpServerAccessLogFilter.java file (line 59), you use the micronaut to retrieve information via request.getPath, request.getMethod, etc.

On the micronaut site (https://docs.micronaut.io/latest/api/io/micronaut/http/HttpRequest.html), I couldn't find a method to retrieve the size of a customer.

Your great experience on micronaut, you may be able to find.

tchiotludo:
It will not be on Request but on [Response](https://docs.micronaut.io/latest/api/io/micronaut/http/HttpResponse.html), I think you could get it from `content-length` headers (if available)

tchiotludo:
@meeraj257 it's assigned for you

rahulkeshervani:
@ericmagna , @tchiotludo I have raised a PR , please review

<img width="1377" alt="Image" src="https://github.com/user-attachments/assets/dfee1032-63f2-4d19-83f4-4dedf19d0ef8" />

ericmagna:
yes thanks to update with size information. It's OK.

thanks.

PR #2116, opened 2025-03-15
fix(log): fix value of content-length always return -1

nicolasdelvat:
Following #1154, the content-lengh displayed is always -1.
This is because the log is evaluated in the OnNext function and this is not the end of the request.
The real end seems to be in "OnComplete".
This also mean the log is now at the end of the request, it was near the end before, so the request time was a little bit off.

Log before : 
image

Log After :
image

Don't hesitate if i'm being too pushy with all the fix PR, I also can create issues if you want. 
I'm just fixing things as I see/need them.

### DIFF
--- a/application.example.yml
+++ b/application.example.yml
@@ -39,7 +39,7 @@ akhq:
     access-log: # Access log configuration (optional)
       enabled: true # true by default
       name: org.akhq.log.access # Logger name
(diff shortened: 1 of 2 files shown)


rahulkeshervani:
@tchiotludo please review this. this is my mess

PR #2125, opened 2025-03-20
chore: do not pass externally mapped groups in token

### LINKED ISSUE #2111: Cookie-size overflow when using external (REST) role mapping
We have implemented a [REST-based role-mapper](https://akhq.io/docs/configuration/authentifications/external.html).

It fails for users who have access to a large number of topics, because [the JWT is rewritten to contain all the mapped groups](https://github.com/tchiotludo/akhq/blob/dev/src/main/java/org/akhq/security/claim/CustomClaimsGenerator.java#L35), which in turn contain a large number of patterns for our topics.

The JWT is further set in a cookie, and cookies are typically [limited to about 4 KiB](https://support.convert.com/hc/en-us/articles/4511582623117-cookie-size-limits-and-the-impact-on-the-use-of-convert-goals#s3). Even though the content is gzipped and BASE64-encoded, the size by far exceeds 4 KiB in our clusters.

We think it would be possible to keep the mapping server-side, and look it up based on the original JWT.

sverrehu:
Closes https://github.com/tchiotludo/akhq/issues/2111

Please see discussion under the above issue.

This PR has four main changes:

1. In `OidcUserDetailsMapper`, do _not_ use the `ClaimProvider` to change the groups in the token, but rather keep the original groups from OIDC. Also, add a `provider_name` attribute as a flag that the groups must be looked up later.
1. In `AKHQSecurityRule` rewrite the `decompressGroups` method (renamed to `unrollGroups`) to check if the flag is present, and in that case use the `ClaimProvider` to map and replace the groups.
1. Change tests in `OidcAuthenticationProviderTest` to use `unrollGroups` before checking the results.
1. Add caching of RestApiClaimProvider response.

In effect, this PR moves the remapping of groups from the OIDC login step, to every request from the user.


src/main/java/org/akhq/security/mapper/OidcUserDetailsMapper.java
```diff
@@ -63,20 +56,10 @@ public Publisher<AuthenticationResponse> createAuthenticationResponse(String pro
...
-
-        try {
-            ClaimResponse claim = claimProvider.generateClaim(request);
-            return (Flowable.just(AuthenticationResponse.success(oidcUsername, List.of(SecurityRule.IS_AUTHENTICATED), Map.of("groups", claim.getGroups()))));
-        } catch (Exception e) {
-            String claimProviderClass = claimProvider.getClass().getName();
-            return Flowable.just(new AuthenticationFailed("Exception from ClaimProvider " + claimProviderClass + ": " + e.getMessage()));
-        }
+        Map<String, Object> attributes = new HashMap<>();
+        attributes.put("provider_name", providerName);
```
AlexisSouquiere, line 61: I think we should add
attributes.put("provider_type", ClaimProviderType.OIDC);
to be able to make the AKHQSecurityRule.getClaimProviderGroups generic if we want to do the same for the other auth providers

src/main/java/org/akhq/security/rule/AKHQSecurityRule.java
```diff
@@ -111,15 +117,44 @@ public Publisher<SecurityRuleResult> check(HttpRequest<?> request, Authenticatio
...
             log.trace("JWT payload is not compressed, returning groups directly");
-            return (Map<String, List<?>>) authentication.getAttributes().get("groups");
+            return authentication.getAttributes().get("groups");
+        }
+    }
+
+    private static Map<String, List<Group>> getClaimProviderGroups(String providerName, List<String> claimGroups, Authentication authentication, ClaimProvider claimProvider) {
+        Map<String, List<Group>> groups = new HashMap<>();
+        ClaimRequest request = ClaimRequest.builder()
+                                           .providerType(ClaimProviderType.OIDC)
```
AlexisSouquiere: And we can use 
providerType(ClaimProviderType.valueOf((String) authentication.getAttributes().get("provider_type")))
instead

src/main/java/org/akhq/security/rule/AKHQSecurityRule.java
```diff
@@ -111,15 +117,44 @@ public Publisher<SecurityRuleResult> check(HttpRequest<?> request, Authenticatio
             return Flowable.just(SecurityRuleResult.REJECTED);
     }
 
-    public static Map<String, List<?>> decompressGroups(Authentication authentication) {
+    public static Map<String, List<Group>> unrollGroups(Authentication authentication, ClaimProvider claimProvider) {
+        Object decompressedGroups = decompressGroups(authentication);
+        if (decompressedGroups instanceof Map<?, ?>) {
+            return (Map<String, List<Group>>) decompressedGroups;
```
AlexisSouquiere: I have a cast issue when I try it with basic auth. The [convert](https://github.com/tchiotludo/akhq/pull/2125/files#diff-064981caf5f79daa76e4d22f89182c40f015cc4bcc12ee4cf3719892cf9f078fR87-R90) has to be moved here

Something like this
```java
    public static List<Group> unrollGroups(Authentication authentication, ClaimProvider claimProvider) {
        Object decompressedGroups = decompressGroups(authentication);

        if ((decompressedGroups instanceof Map<?, ?>)) {
            return ((Map<String, List<?>>) decompressedGroups).values().stream()
                .flatMap(Collection::stream)
                // Type mismatch during serialization from LinkedTreeMap to Group if we use List<Group>
                // Need to serialize Object to Group manually in the stream
                .map(gb -> new ObjectMapper().convertValue(gb, Group.class))
        }

        final String providerName = (String) authentication.getAttributes().get("provider_name");
        if (claimProvider == null || providerName == null || !(decompressedGroups instanceof List<?>)) {
            throw new RuntimeException("No ClaimProvider, or no providerName, or wrong group format");
        }

        return getClaimProviderGroups(providerName, (List<String>) decompressedGroups, authentication,
            claimProvider).values().stream()
            .flatMap(Collection::stream)
    }
```

seems to work

AlexisSouquiere (changes requested):
Appart from the comments that I put to make it more generic and fix the cast issue, your proposal looks good to solve the 4Kb cookie issue for OIDC.
We will be able to do the same for the other providers (except for direct OIDC mapping and the external claim provider as we discussed previously)

src/main/java/org/akhq/security/mapper/OidcUserDetailsMapper.java
sverrehu: Thanks for the review! This one should be fixed now.

src/main/java/org/akhq/security/rule/AKHQSecurityRule.java
sverrehu: This is done as well.

src/main/java/org/akhq/security/rule/AKHQSecurityRule.java
sverrehu: This one fails with ["inference variable R has incompatible bounds"](https://github.com/tchiotludo/akhq/actions/runs/14115829421/job/39545688216?pr=2125#step:9:33). I will have to look into it tomorrow.

sverrehu: I think this is fixed now. Could you PTAL, @AlexisSouquiere ?

AlexisSouquiere: Will do that tomorrow ! I will start to have a look to the caching part too

src/main/java/org/akhq/security/rule/AKHQSecurityRule.java
```diff
@@ -111,15 +114,58 @@ public Publisher<SecurityRuleResult> check(HttpRequest<?> request, Authenticatio
             return Flowable.just(SecurityRuleResult.REJECTED);
     }
 
-    public static Map<String, List<?>> decompressGroups(Authentication authentication) {
+    public static Map<String, List<Group>> unrollGroups(Authentication authentication, ClaimProvider claimProvider) {
```
AlexisSouquiere, line 117: Can you add a comment to explain that the difference between OIDC (groups) and the rest (mapping roles/clusters/patterns) to keep it in mind ?

AlexisSouquiere (changes requested):
Looks good to me !

I'd just need 2 minor changes:
- Add a cache on the LocalSecurityClaimProvider.generateClaim method. Using OIDC with a classic mapping from the config file causes the same behaviour as the RESTAPIClaimProvider (the method is called every time). We can only cache when OIDC is enabled with `@Cacheable(condition = "#{request.providerType == T(org.akhq.models.security.ClaimProviderType).valueOf('OIDC')}")
` (I tested it locally, works fine)
I don't think we need to add in cache for the other authentication mechanism. I will just add data in cache for nothing...

- Add a bit of documentation (maybe [here](https://akhq.io/docs/configuration/authentifications/), to explain that we have 2 caches that people can customise according to their need. (Only used for OIDC, the rest cache when external claim provider is enabled, the local otherwise)

Sounds good for you ?

sverrehu:
> * Add a cache on the LocalSecurityClaimProvider.generateClaim method. Using OIDC with a classic mapping from the config file causes the same behaviour as the RESTAPIClaimProvider (the method is called every time). We can only cache when OIDC is enabled with `@Cacheable(condition = "#{request.providerType == T(org.akhq.models.security.ClaimProviderType).valueOf('OIDC')}") ` (I tested it locally, works fine)

Cache added with your conditional.

sverrehu:
> * Add a bit of documentation (maybe [here](https://akhq.io/docs/configuration/authentifications/), to explain that we have 2 caches that people can customise according to their need. (Only used for OIDC, the rest cache when external claim provider is enabled, the local otherwise)

Added a little bit. :-)

src/main/java/org/akhq/security/rule/AKHQSecurityRule.java
sverrehu: Done.

sverrehu:
Please take another look, @AlexisSouquiere 
Thank you!

AlexisSouquiere (approved):
Ok for me for merging.
Thank you !

reidmeyer:
I'm having an issue after this change. Perhaps you can give me some guidance?

Is there any change in how I should configure my application to conform to this PRs changes?

https://github.com/tchiotludo/akhq/issues/2380

Is the cache now a required configuration to set, for example?

reidmeyer:
> I'm having an issue after this change. Perhaps you can give me some guidance?
> 
> Is there any change in how I should configure my application to conform to this PRs changes?
> 
> #2380
> 
> Is the cache now a required configuration to set, for example?

I was able to get passed the above error by explicitly setting the micronaut cache configuration.

I'm not getting a login loop, along with warns in the logs that say: encodedGroups string exceeds 4 KiB, which is likely to lead to cookie overflow

I'm researching to try to understand what else I might be missing.

Issue #2047, opened 2025-01-23
Bug: Filter with timestamp using start and end timestamp reveals no results if sorted by NEWEST


mauho:
We are using AKHQ 0.25.1 with user settings / Topic Data / Sort = NEWEST

The behavior depends on the size of the topic. Working on a topic containing 40GB of data the following applies:
Setting a start AND end timestamp on a topic won't reveal any data. If however start and end timestamp are set, then sorting is switched to OLDEST, the data is fetched as expected.

If only a start timestamp is used results are fetched even on sorting = NEWEST

```
only start TS   NEWEST   OK  https://address.to/kafka-ui/api/broker-cluster/topic/my-sample-topic/data?sort=NEWEST&partition=All&timestamp=2025-01-06T10%3A25%3A03.000%2B01%3A00
start + end TS  NEWEST   NOK https://address.to/kafka-ui/api/broker-cluster/topic/my-sample-topic/data?sort=NEWEST&partition=All&timestamp=2025-01-06T10%3A25%3A03.000%2B01%3A00&endTimestamp=2025-01-14T10%3A25%3A03.000%2B01%3A00
start + end TS  oldest   OK  https://address.to/kafka-ui/api/broker-cluster/topic/my-sample-topic/data?sort=Oldest&partition=All&timestamp=2025-01-06T10%3A25%3A03.000%2B01%3A00&endTimestamp=2025-01-14T10%3A25%3A03.000%2B01%3A00
```

The behavior is different on smaller topics. Working on a topic containing 1GB of data, the issue is not present -> start+end time filter works regardless of sorting direction.

Can someone try this on a different system before digging into it any deeper?

Thank you and best Regards
Maurizio

AlexisSouquiere:
I think we should prevent the search / newest combination from the UI directly.

- Searching / oldest makes sense because we poll messages until we reach the end or until we found enough matching messages. It's the "normal" way to read a topic.
- Searching / newest doesn't make sense because we need to play with the offset to reverse reading the topic, which is inefficient and won't work on large topics

Automatically setting the sort to oldest in case of search in the UI can do the trick. And maybe warn the user when changing the sort

mikaelol:
We are experiencing the same issue. The problem only arises when we specify an **end timestamp**, if we specify start timestamp (and no end timestamp) and sort by newest it works even for large topics is our experience. I have created a fix here with the suggestions of @AlexisSouquiere : https://github.com/tchiotludo/akhq/pull/2132

[...]

PR #2132, opened 2025-03-25
fix(ui): switch to sort by oldest when specifying timestamp

mikaelol:
Kind of fixing https://github.com/tchiotludo/akhq/issues/2047 (kind of because from the user perspective it should work to sort both ways regardless of topic size).

This PR changes the sort order to `Oldest` every time the user specifies an end timestamp, because it is then the problem arises. And if the user changes to `Newest`, they will be warned that this may not show any data:

image

### DIFF
--- a/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
+++ b/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
@@ -721,14 +721,13 @@ class TopicData extends Root {
         <Dropdown.Item
           key={option}
           onClick={() => {
-              if (option === 'Newest' && this.state.endDatetime !== '') {
-                toast.warn('Sorting by newest with timestamp in large topics may not show data.');
-              }
-              this.setState({ sortBy: option }, () => {
-                this._searchMessages();
-              })
+            if (option === 'Newest' && this.state.endDatetime !== '') {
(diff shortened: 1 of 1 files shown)


mikaelol:
Have you had a chance to review this, @AlexisSouquiere? We've been using this fix in production since March, and it has been working well.

AlexisSouquiere (changes requested):
Sorry I missed this PR. 
Can you run prettier on this file (prettier --write ./src/containers/Topic/Topic/TopicData/TopicData.jsx) ? There are small changes to apply (single quotes instead of double quotes, missing semicolon, etc.)

mikaelol:
> Sorry I missed this PR. Can you run prettier on this file (prettier --write ./src/containers/Topic/Topic/TopicData/TopicData.jsx) ? There are small changes to apply (single quotes instead of double quotes, missing semicolon, etc.)

Fixed now (by running the prettier cmd)

AlexisSouquiere (approved):
Thanks !

svyatos-s21:
Hello, team! We've been experiencing this issue #2047 with blackscreen due to `Uncaught RangeError: Invalid time value` in AKHQ 0.26.0 and latest 0.27.1 versions.

Can this change be applied to the latest or upcoming versions? Thanks!

PR #2134, opened 2025-03-26
feat(ui): make datetime input editable

mikaelol:
This change makes the date-time input field editable, so you are not limited to the 15-minute intervals of the date picker (we are in need of more granular time picking).

image

The trick used here is to use two synchronized date pickers, with one of them not displaying the picker.


erikgb (approved):
LGTM

PR #2138, opened 2025-03-27
Fix UI is broken after clicking empty topic button

### LINKED ISSUE #2043: Bug: ui is broken after clicking empty topic button
**Stop to reproduce:**
Got to topic by clicking on the topics menu on the left side
Open a topic in the topic list
Open data tab
Click empty Topic

**Expected behaviours:**

After clicking empty topic we expect to see data list with no data available message

**Current behaviours**

Empty broken ui with errors messages in the dev console

empty-topic-erros

AmitAnveri:
Fix #2043 

The ref to TopicData was returning null because it is a class component wrapped with withRouter, which doesn't forward refs by default. To fix this, the ref-forwarding logic was implemented in Topic.jsx using a wrapper component that passes the ref as a forwardRef prop through withRouter and applies it to the actual component.

### DIFF
--- a/client/src/containers/Topic/Topic/Topic.jsx
+++ b/client/src/containers/Topic/Topic/Topic.jsx
@@ -66,7 +66,6 @@ class Topic extends Root {
         topicInternal: this.props.location.internal
       },
       () => {
-        console.log('Topic data current:', this.topicData.current);
         this.getTopicsConfig();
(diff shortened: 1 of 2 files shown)


AlexisSouquiere (approved):
Sorry I forgot to review this one.
LGTM, thanks !

PR #2176, opened 2025-05-11
fixed ui is broken after clicking empty topic button (Show "No data available" Message for Empty Topics in Data Tab)

Dev0711:
this PR improves the user experience when viewing the data tab of an empty topic. Previously, if a topic had no messages, the UI would break and display errors in the console. With this update, the data tab now gracefully displays a user-friendly "No data available" message when there are no messages to show. This prevents UI breakage and provides clear feedback to users when a topic is empty.


AlexisSouquiere (commented):
Thank you for the proposal.
The actual version already displays a "No data available" message 
image

The issue is caused by a ref, which is fixed in #2138

AlexisSouquiere:
Closing as fixed by #2138

Dev0711:
Okay, thanks for the reply and inform me about the resolved issue.

Issue #2247, opened 2025-05-20
Can't change default value of 'retention.ms' for a new topic greater than 2,147,483,647


sotik11:
I need to set the default value of 'retention.ms' for the new topic to 315576000000
That's why I wrote the following code in the settings:
```
  topic:
    retention: 315576000000
    partition: 6
    replication: 3
```
But if I set any value greater than 2,147,483,647 (the maximum value of Integer in Java), I get the following error:
`Internal Server Error: Failed to inject value for field [retention] of class: org.akhq.controllers.TopicController Message: Failed to inject value for field [retention] of class: org.akhq.controllers.TopicController Message: Error resolving property value [${akhq.topic.retention}]. Unable to convert value [31557600000] to target type [Integer] due to: For input string: "315576000000" Path Taken: new TopicController() --> TopicController.retention Path Taken: new TopicController() --> TopicController.retention`
And at the same time when I am creating a topic and enter, for example, the number 315576000000 (10 years in milliseconds) manually, then everything is without errors.
Can you fix it please?

![Image](https://github.com/user-attachments/assets/439aae4c-ee9d-4f23-8eff-6e249e59f805)

iddeepak:
The root cause here is that TopicController here is binding akhq.topic.retention into an Integer, whose max is 2 147 483 647. Any value above 2,147,483,647 simply can’t be parsed by Spring, even though Kafka itself uses a 64-bit retention:
[See this Line](https://github.com/apache/kafka/blob/896f283e1e5e9d6ce4d4a7591d2654d74c1ebf1a/storage/src/main/java/org/apache/kafka/storage/internals/log/LogConfig.java#L71)

Not sure why AHQ uses an Integer. We should switch the controller field (and API) to long/Long so you can set any realistic retention without hitting a limit.

PR #2249, opened 2025-05-21
 Fix: Support Large Retention Values in Topic Config

iddeepak:
The [Issue:2247](https://github.com/tchiotludo/akhq/issues/2247) was caused because TopicController.retention was defined as an Integer, which has a max value of 2,147,483,647 — too small for realistic retention.ms values like 315576000000 (10 years in milliseconds).

Since Kafka supports this config as a 64-bit long, the field should be updated to use Long instead of Integer.

### DIFF
--- a/src/main/java/org/akhq/controllers/TopicController.java
+++ b/src/main/java/org/akhq/controllers/TopicController.java
@@ -79,10 +79,10 @@ public class TopicController extends AbstractController {
 
     @Get ("api/topic/defaults-configs")
     @Operation(tags = {"topic"}, summary = "Get default topic configuration")
-    public Map<String,Long> getDefaultConf(){
+    public Map<String,Object> getDefaultConf(){
         return Map.of(
-            "replication", replicationFactor.longValue(),
-            "partition", partitionCount.longValue(),
+            "replication", replicationFactor,
+            "partition", partitionCount,
             "retention", retention
         );
     }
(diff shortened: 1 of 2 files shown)


src/main/java/org/akhq/controllers/TopicController.java
```diff
@@ -73,16 +73,16 @@ public class TopicController extends AbstractController {
...
     private Integer pageSize;
 
     @Get ("api/topic/defaults-configs")
     @Operation(tags = {"topic"}, summary = "Get default topic configuration")
-    public Map<String,Integer> getDefaultConf(){
+    public Map<String,Long> getDefaultConf(){
         return Map.of(
-            "replication", replicationFactor.intValue(),
-            "partition", partitionCount,
+            "replication", replicationFactor.longValue(),
```
AlexisSouquiere: I think it would be better to keep the original type instead of converting replication and partition to long. Can you change the return type to Map<String, Object> and remove the `.longValue()` please ?

AlexisSouquiere (changes requested):
Please take into account my comment.
Otherwise it's ok for me. Thank you !

src/main/java/org/akhq/controllers/TopicController.java
iddeepak: Addressed as per the suggestion. Thank you @AlexisSouquiere

AlexisSouquiere (approved):
LGTM, thank you !

PR #2252, opened 2025-05-23
delete check topicname againts cgroups patterns

### LINKED ISSUE #2251: Topic consumer groups tab not working properly with regex
The consumer groups tab inside the topic page is not working properly when using regex filters for consumer groups. The user has permissions over the topic and all its consumer groups, but when entering the tab, an AuthorizationException is thrown and the user is automatically logged out.

That is because of the tag @AKHQSecured. The configured resource is CONSUMER_GROUP , but then the method _checkIfClusterAndResourceAllowed_ is called with the topic name. That method will check the topic name against the regex filters set in the consumer groups, so it will fail.

    @AKHQSecured(resource = Role.Resource.CONSUMER_GROUP, action = Role.Action.READ)
    @Get("api/{cluster}/topic/{topicName}/groups")
    @Operation(tags = {"topic"}, summary = "List all consumer groups from a topic")
    public List<ConsumerGroup> groups(String cluster, String topicName,
                                      Optional<TopicRepository.TopicGroupsListView> groupsListView)
        throws ExecutionException, InterruptedException {
        checkIfClusterAndResourceAllowed(cluster, topicName);

The proposed fix would be not to check the topic resource along with the consumer group resource set in the tag.

alozano3:
This PR fixes #2251

### DIFF
--- a/src/main/java/org/akhq/controllers/TopicController.java
+++ b/src/main/java/org/akhq/controllers/TopicController.java
@@ -278,7 +278,7 @@ public class TopicController extends AbstractController {
     public List<ConsumerGroup> groups(String cluster, String topicName,
                                       Optional<TopicRepository.TopicGroupsListView> groupsListView)
         throws ExecutionException, InterruptedException {
-        checkIfClusterAndResourceAllowed(cluster, topicName);
+        checkIfClusterAllowed(cluster);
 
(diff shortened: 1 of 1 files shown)


AlexisSouquiere (approved):
Good catch. Check on the cluster is enough. Filter on the consumer groups is done afterwards in the `buildUserBasedResourceFilters`

PR #2253, opened 2025-05-24
Fix consumer group auth check in topic groups endpoint

White-XA:
Topic name was incorrectly being checked as a consumer group resource.
This PR replaces `checkIfClusterAndResourceAllowed(cluster, topicName)` with `checkIfClusterAllowed(cluster)`.


AlexisSouquiere:
Thank you for the contribution. This issue is already managed by #2252

PR #2293, opened 2025-06-10
feat(helm): add configurable targetPort support for sidecars

kiambogo:
This PR updates the Helm charts to support a configurable service targetPort. The current implementation has the ingress hardcoded to the `http` named port, and the service will point to an `http` named port in the deployment. K8s sidecars do not support named ports being aggregated under a Service ([ref](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior)), so having these named ports hardcoded today prevents the ability of deploying/configuring AKHQ to use a reverse proxy sidecar (e.g. when using header-auth in AKHQ). 

- Add configurable service.targetPort and service.managementTargetPort options
- Update service template to support custom target ports with defaults
- Update ingress template to use configurable targetPort

### DIFF
--- a/helm/akhq/templates/ingress.yaml
+++ b/helm/akhq/templates/ingress.yaml
@@ -49,7 +49,7 @@ spec:
               service:
                 name: {{ $fullName }}
                 port:
-                  name: http
+                  name: {{ .Values.service.targetPort | default "http" }}
               {{ else }}
               serviceName: {{ $fullName }}
               servicePort: http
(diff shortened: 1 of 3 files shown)


helm/akhq/templates/ingress.yaml
```diff
@@ -49,7 +49,7 @@ spec:
               service:
                 name: {{ $fullName }}
                 port:
-                  name: http
+                  name: {{ .Values.service.targetPort | default "http" }}
```
penguinoneshaw, line 52: @tchiotludo This has introduced a fairly serious error in the chart - it's missing the `$` at the beginning to anchor to the values from the root so it tries to get `.Values` of a string. This means trying to use the chart gets the error

```
can't evaluate field Values in type interface {}
```

tchiotludo: you can create a PR to fix?

PR #2406, opened 2025-08-05
fix(auth): use configured context-path in error controller

### LINKED ISSUE #2405: ErrorController does not account for configured context path when checking request URI
When running AKHQ that uses a configured `context-path` (e.g. `/kafka-ui`), the logic in `ErrorController` fails to correctly detect API routes due to a hardcoded URI prefix check:

```java
if (request.getUri().toString().startsWith("/api")) {
    ...
}
```

Since request.getUri().toString() includes the full path (including the context path), this condition will not match when the application is deployed under a sub-path like /kafka-ui. For example, the actual URI might be /kafka-ui/api/..., causing the check to fail even though the route is valid.

## Suggested Fix

```java

if (request.getUri().toString().startsWith(getBasePath()+ "/api")) {
    ...
}
```

doxsch:
closes #2405


doxsch:
Hello @tchiotludo,

would it be possible to include this in v0.27.1?
I was hoping it would already make it into the v0.27.0 release.

Thank you! 🙏

AlexisSouquiere:
Hi @doxsch, did you test the fix on your side? I think there is something additional to fix because `BasicHttpAttributes.getRouteMatchInfo(request).isPresent()` won't match so the `ErrorController` doesn't return the forbidden error. But the thing is that I'm experiencing the same with or without the `context-path`. 

It has been a long time since I worked on this, I need to investigate further

AlexisSouquiere:
Just reviewed another PR linked to this problem. 
#2858 will fix it

doxsch:
> Just reviewed another PR linked to this problem. #2858 will fix it

These changes are still needed so that the behavior is the same with and without a context path.

AlexisSouquiere (approved):

PR #2420, opened 2025-08-18
fix(ui): preserve precision for large integer configuration values

### LINKED ISSUE #2402: Large 64-bit integer configuration values are rounded, probably due to limited Javascript number precision
The default value for many settings are large 64-bit integer values, and AKHQ user interface incorrectly rounds these.  It looks like the back-end sends them correctly as JSON strings, so that the number isn't rounded.  But then it looks like the user interface decides to parse the number, and then we end up with an incorrect value being shown.

The right solution might be for the UI to not do any number parsing, and treat them as strings...??

See the screenshot below:

Image

Expected behavior in the screenshot would be for the configuration setting to show `9223372036854775807`.  Instead, it shows `9223372036854776000`.

This behavior was observed in AKHQ 0.26.0.

mbaurin:
Hi @tchiotludo,

This PR fixes a critical precision issue in the frontend where large 64-bit integer configuration values were being incorrectly displayed due to JavaScript's number precision limitations.

## Problem

When displaying Kafka configuration values in Topic and Node configuration pages, large integers like `9223372036854775807` (Long.MAX_VALUE) were losing precision and showing as `9223372036854776000`. This happened because the frontend was converting JSON string values to  JavaScript numbers using `+config.value`, which exceeds JavaScript's `Number.MAX_SAFE_INTEGER` (9007199254740991).

## Solution

- **Created utility functions** (`utils/largeIntegerUtils.js`) to detect and safely handle large integers
- **Fixed TopicConfigs and NodeConfigs components** to preserve large integers as strings while maintaining backward compatibility for normal numeric values
- **Added comprehensive test coverage** (19 tests) including the exact problematic value from the issue
- **Zero breaking changes** - existing functionality preserved for all normal configuration values

## Key Changes

- `safeConfigValue()` replaces `+config.value` logic to conditionally preserve large integers as strings
- ConnectConfigs already handled this correctly (no changes needed)
- Uses existing `lossless-json` dependency (already in package.json)
- Follows project's React class component patterns and coding conventions

## Testing
- All new utility function tests pass
- No regression in existing functionality  

Closes #2402

### DIFF
--- a/client/src/containers/Node/NodeDetail/NodeConfigs/NodeConfigs.jsx
+++ b/client/src/containers/Node/NodeDetail/NodeConfigs/NodeConfigs.jsx
@@ -4,6 +4,7 @@ import { BYTES, MILLI, TEXT } from '../../../../utils/constants';
 import Table from '../../../../components/Table';
 import Form from '../../../../components/Form/Form';
 import converters from '../../../../utils/converters';
+import { safeConfigValue } from '../../../../utils/largeIntegerUtils';
 import Joi from 'joi-browser';
 import { toast } from 'react-toastify';
 import 'react-toastify/dist/ReactToastify.css';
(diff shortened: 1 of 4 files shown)


AlexisSouquiere (approved):
LGTM, thank you !

Issue #2499, opened 2025-09-25
Capability to disable a single controller method by configuration


fabfreitas:
Hello, 

I would like to know if it is possible add the capability to disable a single controller method by configuration, like as example below:

```
controllers:
  mycontroller:
    hello:
       enabled: false
```
and...

```
import io.micronaut.context.annotation.Requires;
import io.micronaut.http.annotation.Controller;
import io.micronaut.http.annotation.Get;

@Controller("/mycontroller")
public class MyController {

    @Requires(property = "controllers.mycontroller.hello.enabled", value = "true")
    @Get("/hello")
    public String index() {
        return "Hello, World!";
    }
}
```

The reason is I would like to configure some api endpoints to not be reachable, while others will be.

Thank you in advance

AlexisSouquiere:
@fabfreitas using the new RBAC isn't enough for you ? What is your need behind that ?
Some endpoints rely on other endpoints during navigation in the UI. For instance on the topics screen, you need /topic, /groups/topics, /topic/last-record. IMO, it would introduce potential issues setting up this behaviour if you don't know exactly how it works

PR #2507, opened 2025-10-02
feat: add controller method toggle via configuration

rahghav:
## Description

  Implements the capability to disable individual controller endpoints via configuration, as requested in #2499 by
  @fabfreitas.

  This PR adds a flexible mechanism to enable/disable specific controller methods without modifying code, allowing
  administrators to restrict access to certain API endpoints based on deployment requirements.

  ## Changes

  - **Added `@ControllerMethodEnabled` annotation**: Custom annotation to mark controller methods as toggleable
  - **Added `ControllerMethodEnabledCondition` class**: Micronaut condition that evaluates configuration properties at
  startup
  - **Updated `TopicController`**: Applied annotation to `create()` method as a demonstration
  - **Updated `application.example.yml`**: Added configuration examples
  - **Added `CONTROLLER_METHOD_TOGGLE.md`**: Comprehensive documentation with usage examples

  ## Usage Example

  ### In Controller:
  ```java
  @ControllerMethodEnabled(property = "akhq.controllers.topic.create.enabled", defaultValue = true)
  @Post(value = "api/{cluster}/topic")
  public Topic create(...) {
      // method implementation
  }

  In Configuration:

  akhq:
    controllers:
      topic:
        create:
          enabled: false  # Disables the topic creation endpoint

  Benefits

  - Security: Easily disable sensitive operations in production environments
  - Flexibility: Different configurations for different environments without code changes
  - Granular Control: Enable/disable individual endpoints rather than entire controllers
  - No Code Changes Required: Toggle features through configuration only

  How It Works

  The implementation uses Micronaut's @Requires condition system. When an endpoint is disabled via configuration, it is
  not registered at application startup and will return 404 when accessed.

  Testing

  The feature has been demonstrated on TopicController.create() method. To test:

  1. Add configuration to disable an endpoint
  2. Start the application
  3. Verify the endpoint returns 404

### DIFF
--- a/CONTROLLER_METHOD_TOGGLE.md
+++ b/CONTROLLER_METHOD_TOGGLE.md
@@ -1,0 +1,129 @@
+# Controller Method Toggle Feature
+
+## Overview
+
+This feature allows you to enable or disable specific controller endpoints via configuration without modifying the code. This is useful for restricting access to certain API endpoints in specific environments or deployments.
+
+## Usage
+
+### 1. Annotate Controller Methods
+
+Add the `@ControllerMethodEnabled` annotation to any controller method you want to make toggleable:
+
+```java
+@Controller
+public class TopicController extends AbstractController {
+
+    @AKHQSecured(resource = Role.Resource.TOPIC, action = Role.Action.CREATE)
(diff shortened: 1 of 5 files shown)


rahghav:
@fabfreitas

Issue #1501, opened 2023-06-08
Fail to load javascript when expanding topic message data


Adman:
Hello,

After clicking on topic message data to expand it, I'm getting the following error message in developer console:
```
Refused to execute script from 'https://<my-domain>/kafka/ui/ui/33be477618f676447f25a3fa5252f3d0.js' because its MIME type ('text/html') is not executable.
```
The `context-path` is set to `"/kafka"`.
Most probably, the problem is that it's adding one more "ui" string into the url. If I remove one of the "ui" from the URL, it works.

napalmCore:
Hi, i would like to help, could you assign me this?

napalmCore:
Hi.
I was not able to reproduce the bug, i cloned the project and run the docker-compose-dev the project is built and run successfully,
When I go to the topic page and open the data tab I don't see any errors except a last pass extension related error.
I tested it on chrome 131.0.6778.265 and Edge 131.0.2903.112
See the screenshot attached
Could you please provide more context on how to reproduce this bug
![topics-data-ui-no-errors](https://github.com/user-attachments/assets/809db998-ec29-4b35-a3af-fba96db29549)

Adman:
@napalmCore Try sending a json message to the topic and then expand it.

napalmCore:
@Adman do you mean add JSON to the value field in create topic form? 
![add-topic](https://github.com/user-attachments/assets/ba12cece-54d0-4bc3-9488-6b3074c5030b)

Adman:
Yes. Or create for example python client which is able to send that message there.

napalmCore:
@Adman unfortunately i tried both methods and it works fine, i was not able to reproduce the bug. i tried even with a big JSON with nested data.

hitdepani:
I WOULD LIKE TO WORK ON THIS

PR #2586, opened 2025-11-06
1 fix(client/tail): Display message headers correctly in Live Tail (Issue #1501)

swayamraut8:
The Issue: The message headers were not being displayed in the tail view. The "Headers" column was not correctly rendering the headers when expanded because the code was trying to iterate over an object as if iit were an array.
  
The Fix:

   1. Correct Header Iteration: In client/src/containers/Tail/Tail.jsx, I modified the onExpand function to
      use Object.entries(obj.headers). This converts the headers object into an array, allowing us to
      correctly iterate over and display each header's key and value.

   2. Improved Readability: The headers are now displayed in a structured, nested table within the expanded
      row, making them easier to read.

   3. Enhanced UI: I've added descriptive labels such as "Headers" and "Value" to make the interface more
      user-friendly.

  These changes ensure that the message headers are now correctly and clearly displayed in the tail view,
  resolving the issue.

### DIFF
--- a/client/src/containers/Tail/Tail.jsx
+++ b/client/src/containers/Tail/Tail.jsx
@@ -462,15 +462,15 @@ class Tail extends Root {
                 cell: obj => {
                   return (
                     <div className="tail-headers">
-                      {obj.headers ? Object.keys(obj.headers).length : 0}
+                      Headers ({obj.headers ? Object.keys(obj.headers).length : 0})
                     </div>
                   );
                 }
               },
               {
-                id: 'value',
+                id: 'Value',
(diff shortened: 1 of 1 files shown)


swayamraut8:
The

AlexisSouquiere:
@swayamraut8 can you explain me the issue you are facing ?
I did a quick test on the dev branch and displaying the headers works for me

image

AlexisSouquiere (changes requested):
I did a quick test on the dev branch and displaying the headers works for me. What is the issue?

PR #2648, opened 2025-12-10
Implemented Schema-Registry Search Filtering by Id, Subject, Version and Schema Type

### LINKED ISSUE #2450: Schema registry search by schema id
Hi,
We are in version 0.25.1

In the 'Schema Registry' section, if we search by schema name, schemas are list with there id and version,
but if we search by id the response is 'No schemas available'

ameen-pj:
Fixes #2450 

Added UI for filtering schemas in the Schema Registry list by Id, Subject, Version and Schema Type
1. Added a `<select>` tag for the dropdown.
2. Implemented filtering logic based on filter type

image

### DIFF
--- a/client/src/containers/Schema/SchemaList/SchemaList.jsx
+++ b/client/src/containers/Schema/SchemaList/SchemaList.jsx
@@ -17,6 +17,13 @@ import Root from '../../../components/Root';
 import { handlePageChange, getPageNumber } from './../../../utils/pagination';
 import { withRouter } from '../../../utils/withRouter';
 
+const FilterType = {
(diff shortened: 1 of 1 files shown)


AlexisSouquiere (changes requested):
Thank you for the proposal @ameen-pj.
For me the filtering shouldn't be done on the frontend side. For instance, when a user searches for schema ID 99, the frontend should call the `api/{cluster}/schema/id/{id}` endpoint to retrieve only all the subjects matching the schema ID.

Devansh-jpg:
Hey is this issue solved?

PR #2668, opened 2025-12-15
Fix schema repository

webcreator30:
# Problem 
Apicurio registry : No version '2' found for artifact with ID 'com.company.movie' in group 'null'.; error code: 40402

# Solution
In the Schema Registry Repository the method toSchemasLatestVersion was not handling exception properly , so added log warning messages 

# Result 
AKHQ portal remains functional even when some schema versions are not present


AlexisSouquiere (changes requested):
Please remove all the updates to the Helm files and Dockerfile

webcreator30:
I’ve removed the Helm and Dockerfile changes and pushed the update. Please let me know if anything else is needed

PR #2745, opened 2026-02-09
Fix: Exclude conflicting Confluent dependencies to resolve Jackson co…

### LINKED ISSUE #2743: Investigate replacement of org.sourcelab:kafka-connect-client
We are using "org.sourcelab:kafka-connect-client:4.0.5" which causes compatibility issues with jackson library (https://github.com/SourceLabOrg/kafka-connect-client/issues/97). We can't update jackson to a latest version because of this.

We should investigate how to get rid of this library (doesn't seem to be maintained anymore). Any alternative available ?

debika-samanta:
…mpatibility issues

- Exclude Jackson from kafka-connect-client to prevent version conflicts
- Exclude proprietary Confluent artifacts (observability, resourcemanager) from ksqldb-api-client
- Add same exclusions to test dependencies (connect-runtime, ksqldb-rest-app)
- Resolves dependency resolution failures with missing Confluent proprietary dependencies

Fixes #2743


AlexisSouquiere (changes requested):
It is not the right approach. The sourcelab kafka connect client refers to `PropertyNamingStrategy.SNAKE_CASE` 

https://github.com/SourceLabOrg/kafka-connect-client/blob/bd3b919ff50587feaf7ceecaa1eb792693808e24/src/main/java/org/sourcelab/kafka/connect/apiclient/request/JacksonFactory.java#L51

It has been removed
https://github.com/FasterXML/jackson-databind/blob/16187a7ecf69fda4eacfab742ab74b07bd3d96d2/src/main/java/com/fasterxml/jackson/databind/PropertyNamingStrategy.java#L54

We can't use the latest jackson version anyway

poonam-meghnani:
I am facing this issue using kafka-connect-client with newer versions of springboot. Is this fixed? which version is this planned to be part of?

AlexisSouquiere:
Fixed with #2967

PR #2827, opened 2026-03-14
fix: include null keys in NOT_CONTAINS filter results

### LINKED ISSUE #2775: Weird behaviour when searching with not contains
Hi,
I was testing some behaviours in AKHQ and noticed that when filtering by whichever one field on Not Contains, then the search does not show all relevant results. For example, without any filters I get results from partitions 4, 9, 16, 21. But when adding such a negative filter, then I get results only from partitions 4 and 16, even if adding just a key filter not contains on a key from partition 4 for example. If I add a key filter contains and something else not contains then it seems to behave as expected.

Anyone can take a look?

Thanks!
Br,
Richard

AbdulHaseeb22:
Fixes #2775

Hey, I looked into this and managed to reproduce it,
I created a topic with 4 partitions and produced some messages 
with null keys, Without any filter all 4 partitions show up fine. 
But as soon as I apply a NOT_CONTAINS key filter, all the null-key 
records just disappear from the results.

The issue is in 'notContainsAll' there's a 
'.filter(Objects::nonNull)' call that silently drops null values 
before the filter even runs. So when a record has a null key, 
the stream is empty and 'anyMatch' just returns false, meaning 
the record gets excluded even though null obviously doesn't 
contain anything.

Fix is pretty simple, removed the null filter and added 
's == null' check directly in the 'anyMatch' lambda, so If the key 
is null it just passes through, which is the correct behaviour.

Tested before and after, null-key records now show up correctly 
with NOT_CONTAINS filter applied.

### DIFF
--- a/src/main/java/org/akhq/repositories/RecordRepository.java
+++ b/src/main/java/org/akhq/repositories/RecordRepository.java
@@ -874,8 +874,7 @@ public class RecordRepository extends AbstractRepository {
         }
 
         return in.parallelStream()
-            .filter(Objects::nonNull)
-            .anyMatch(s -> extractSearchPatterns(search)
+            .anyMatch(s -> s == null || extractSearchPatterns(search)
                 .stream()
                 .noneMatch(s.toLowerCase()::contains));
     }


AlexisSouquiere (approved):

AlexisSouquiere:
@AbdulHaseeb22 good catch, thank you !

PR #2828, opened 2026-03-14
fix(topicdata): use beginningOffset as lower bound in getLastRecord seek

### LINKED ISSUE #2490: get /api/{cluster}/topic/last-record fails when topic has only a single record
The current implementation of `getLastRecord` in `RecordRepository` only works if a topic has more than one record. For topics only one record there is, no result because the code does `endOffset - 2` and then performs a single poll. If the partition only contains one record, this logic skips it.

**Steps to Reproduce:**

- Create a topic with only one partition.
- Produce a single message to that topic.
- Call get /api/{cluster}/topic/last-record for said record
- Empty response returned

AbdulHaseeb22:
Fixes #2490

Hey, I reproduced this by producing 10 records to a topic then 
deleting the first 9, leaving a single record at offset 9.

The issue is in getLastRecord method, it seeks to Math.max(0, endOffset - 2)  which protects against negative offsets but not against going below , the actual first available offset. With the record at offset 9, it seeks to offset 8 which no longer exists, so thats why the poll returns nothing
Fix: fetch the beginning offsets first and use Math.max(beginOffset, 
endOffset - 2) instead, This makes sure we never seek below the first 
available record regardless of how many records were deleted or 
compacted away.

i tested this before and after, empty {} with a single high-offset record 
now correctly returns that record.

### DIFF
--- a/src/main/java/org/akhq/repositories/RecordRepository.java
+++ b/src/main/java/org/akhq/repositories/RecordRepository.java
@@ -106,10 +106,12 @@ public class RecordRepository extends AbstractRepository {
         try (KafkaConsumer<byte[], byte[]> consumer = kafkaModule.getConsumer(clusterId)) {
             consumer.assign(topicPartitions);
 
+            Map<TopicPartition, Long> beginOffsets = consumer.beginningOffsets(consumer.assignment());
             consumer
                 .endOffsets(consumer.assignment())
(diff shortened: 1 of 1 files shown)


AlexisSouquiere (approved):

AlexisSouquiere:
@AbdulHaseeb22 thank you for the PR and for the explanation details !

PR #2856, opened 2026-03-25
feat(connect): add status filter to connect definitions

flysen:
- Introduced a status filter for connect definitions in the ConnectList component.
- Updated the API to support filtering by status in the ConnectController and ConnectRepository.
- Added corresponding tests to validate the new filtering functionality.
- Enhanced settings to manage task status filters.
- Updated constants to include task status options.

changes-kc


AlexisSouquiere:
Hi @flysen ! Thank you for the PR. It looks good to me. 
There is just one thing I'd need regarding the update in the settings. What is the need behind this ? Why would you need to uncheck some of the statuses ?
To me, I'd like to see all the statuses in the dropdown

flysen:
Hi @AlexisSouquiere 
I have 29 pages in pagination, lots of connectors and like to have a quick sort function.
Screenshot 2026-06-15 at 12 50 19

AlexisSouquiere:
Ok so if you're ok, let's split the work in 2. I think we should keep only the status filter in this PR (without the settings part)
Then, we can init a discussion topic to see how we can enhance the connectors list page, especially when we have a huge number of connectors (other search fields? dropdown to choose the number of result?, some sort field?)

Sounds good ?

flysen:
Hi @AlexisSouquiere! I've addressed your feedback — removed the Settings integration so all task statuses (RUNNING, FAILED, PAUSED, UNASSIGNED) are now always shown statically in the dropdown. Also added a tooltip to clarify that the filter shows connectors with at least one task in the selected state. Let me know if this looks good to you!

Screenshot 2026-06-15 at 20 52 08

AlexisSouquiere (approved):

AlexisSouquiere:
Thank you ! Feel free to open a discussion to talk about other enhancements (filters, etc.) we could do on the connectors page 🚀

PR #2857, opened 2026-03-25
fix: poll for task state changes after connector actions

flysen:
After performing a pause, resume, or restart action on a connector,
the task states in the detail view weren't reflecting the new state
until **the user manually refreshed the page**. This was caused by the
Kafka Connect API returning before the state change had fully
propagated across all tasks.

Now polls the target connector's detail endpoint until all its tasks
reach the expected state. A loading spinner is shown while polling,
and if any request fails, an error toast is displayed and the view
refreshes with the latest available state.

Polling is configured via static class constants in ConnectTasks:
- **POLL_INTERVAL**: time between attempts in ms (default: 1000)
- **POLL_MAX_ATTEMPTS**: maximum number of retries (default: 10)


flysen:
Claude review found and [corrected](https://github.com/tchiotludo/akhq/pull/2857/changes/7d757244e3bfe0c1f24eca42fc8dd8347597f91d):

1. **Individual task restart** — Instead of polling all tasks, does a single delayed refresh when restarting one specific task (since waiting for ALL tasks to be RUNNING doesn't make sense for a single task restart)
2. **FAILED as a settled state** — When resuming/restarting, treats FAILED as settled alongside RUNNING. Previously it would poll for 10 seconds if a task went to FAILED, waiting for RUNNING that would never come.
3. **Removed the else { settled = true } fallback** — Unknown actions no longer silently skip polling. If the action doesn't match paused/resumed/restarted, it falls through to the exhausted-attempts refresh instead.

Looks legit and good

AlexisSouquiere (approved):

PR #2858, opened 2026-03-25
fix: return 403 instead of 307 for unauthorized API requests

### LINKED ISSUE #2854: [0.27.0] Read-only user gets false success message when producing to topic (307 instead of 403)
After upgrading from 0.26.0 to 0.27.0, when a read-only user (no `TOPIC_DATA/CREATE` permission) tries to produce a message via "Produce to Topic", the UI shows a success message "Produced to dummy-topic", but the message is never actually produced.

In 0.26.0 the same action correctly returned a 403 and the user was redirected to the login page with an error message.

**Server logs (same user, same endpoint):**

0.26.0 correct behavior:
```
2026-03-24 11:53:56,339 WARN io-executor-thread-4 org.akhq.log.access [Url: POST /api/kafka-cluster/topic/test-topic/data] [Status: 403] [User: TEST] [Length: 87]
```

0.27.0 broken:
```
2026-03-24 12:07:15,334 INFO virtual-executor-967955864 org.akhq.log.access [Url: POST /api/kafka-cluster/topic/test-topic/data] [Status: 307] [User: TEST] [Length: 0]
```
 
I guess the Micronaut upgrade (4.3.8 -> 4.10.7) changed something around route matching, causing `BasicHttpAttributes.getRouteMatchInfo(request)` in the `ErrorController` to return `Optional.empty()`. This would make the code fall through to the `temporaryRedirect` fallback instead of returning a 403

arslanerdemm:
When an AuthorizationException was thrown for an request the ErrorController attempted to look up the original matched route via `BasicHttpAttributes.getRouteMatchInfo()` to read `@AKHQSecured` annotation values for a detailed 403 response. But by the time the global error handler runs, the route match info is no longer present on the request (possibly due to changes in newer Micronaut version) so the lookup always returned empty. This caused the handler to fall through and return a 307 Redirect instead of a proper 403 forbidden.

**Changes:**
- `AKHQSecurityRule` now stores the rejected resource and action as request attributes before returning REJECTED
- `ErrorController` reads those attributes to build a 403 response body, removing the broken route match lookup

Fixes #2854 #2405

### DIFF
--- a/src/main/java/org/akhq/controllers/ErrorController.java
+++ b/src/main/java/org/akhq/controllers/ErrorController.java
@@ -2,8 +2,6 @@ package org.akhq.controllers;
 
 import io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientException;
 import io.micrometer.core.instrument.util.StringUtils;
-import io.micronaut.core.annotation.AnnotationValue;
-import io.micronaut.http.BasicHttpAttributes;
 import io.micronaut.http.HttpRequest;
 import io.micronaut.http.HttpResponse;
 import io.micronaut.http.HttpStatus;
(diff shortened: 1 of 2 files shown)


arslanerdemm:
@AlexisSouquiere would appreciate a review when you get a chance :)

AlexisSouquiere:
Yes, it is in my todo 😄

AlexisSouquiere (changes requested):
Thanks for the PR @arslanerdemm, it's clear and fixes the issue.
Can you just extract the 2 new attributes name in 2 constants in the AKHQSecurityRule ?

arslanerdemm:
> Thanks for the PR @arslanerdemm, it's clear and fixes the issue. Can you just extract the 2 new attributes name in 2 constants in the AKHQSecurityRule ?

@AlexisSouquiere done, thanks for the review!

AlexisSouquiere (approved):

PR #2825, opened 2026-03-13
chore(deps): upgrade to Java 25

AlexisSouquiere:
Close #2824

PR #2884, opened 2026-03-30
fix(build): update remaining Java 17 references after Java 25 upgrade

arslanerdemm:
Follow-up to #2825


AlexisSouquiere:
Good catch, thank you 🚀

PR #2885, opened 2026-03-31
Allow empty config values to fallback to cluster defaults

### LINKED ISSUE #2601: Allow Cleanup Policy & Retention to Be Empty to Use Cluster Defaults During Topic Creation
### **Describe**

When creating a new topic in AKHQ, certain fields—specifically: _Cleanup Policy_ and _Retention_ are currently required or always filled with explicit values.

However, in Kafka, these configurations are optional and can be omitted so that the topic inherits the cluster-level default settings (as defined in the broker configs such as log.cleanup.policy, log.retention.ms, log.retention.bytes, etc.).

### **Requested**

During Topic Creation:
1. Allow Cleanup Policy to be empty, Kafka will use broker-level log.cleanup.policy
2. Allow Retention to be empty, if left empty, AKHQ should not send these configs. Kafka will apply log.retention.ms or log.retention.bytes defaults

### **Why this is useful**

- Aligns AKHQ with Kafka’s native behavior
- Prevents forcing users to define values unnecessarily
- Avoids conflicts between AKHQ-specified configs and broker-managed governance
- Makes topic creation simpler and safer

yunasora:
Fixes #2601 

When creating a new topic, if a user leaves `cleanup.policy` or `retention` blank (intending to use the cluster's default settings), the process previously failed. This PR fixes the issue across both the frontend and backend.

**Changes made:**
1. **Backend:** Added a `.filter()` in the controller to remove empty configuration values (`""`) from the request payload, preventing Kafka from rejecting them.
2. **Frontend:** Updated the Joi validation schema in `TopicCreate.jsx` to `.allow('')` for the retention field, ensuring the UI doesn't block the submission when the field is cleared.

**Testing:**
- Verified locally that leaving Retention blank successfully creates the topic using the cluster default configuration.
- Passed all CI tests.

### DIFF
--- a/client/src/containers/Topic/TopicCreate/TopicCreate.jsx
+++ b/client/src/containers/Topic/TopicCreate/TopicCreate.jsx
@@ -38,7 +38,7 @@ class TopicCreate extends Form {
     partition: Joi.number().min(1).label('Partition').required(),
     replication: Joi.number().min(1).label('Replication').required(),
     cleanup: Joi.string().required(),
-    retention: Joi.number().label('Retention')
+    retention: Joi.number().allow('').label('Retention')
   };
 
   onCleanupChange = value => {
--- a/src/main/java/org/akhq/controllers/TopicController.java
+++ b/src/main/java/org/akhq/controllers/TopicController.java
@@ -149,6 +149,7 @@ public class TopicController extends AbstractController {
             (configs != null ? configs : ImmutableMap.<String, String>of())
                 .entrySet()
                 .stream()
+                .filter(r -> r.getValue()!=null&&!r.getValue().isEmpty())
                 .map(r -> new Config(r.getKey(), r.getValue()))
                 .collect(Collectors.toList())
         );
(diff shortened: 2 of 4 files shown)


AlexisSouquiere (changes requested):
Please revert your changes on the package-lock.json and application.yml files

src/main/java/org/akhq/controllers/TopicController.java
```diff
@@ -149,6 +149,8 @@ public Topic create(
             (configs != null ? configs : ImmutableMap.<String, String>of())
                 .entrySet()
                 .stream()
+                // Triggering GitHub Actions for testing #2601
```
AlexisSouquiere, line 152: Please remove this comment

src/main/java/org/akhq/controllers/TopicController.java
```diff
@@ -149,6 +149,8 @@ public Topic create(
...
+                .filter(r -> r.getValue()!=null&&!r.getValue().isEmpty())
```
AlexisSouquiere, line 153: Please reformat this line

client/src/containers/Topic/TopicCreate/TopicCreate.jsx
```diff
@@ -38,7 +38,7 @@ class TopicCreate extends Form {
     partition: Joi.number().min(1).label('Partition').required(),
     replication: Joi.number().min(1).label('Replication').required(),
     cleanup: Joi.string().required(),
```
AlexisSouquiere, line 40: allow('') should be there too to allow not sending cleanup policy

PR #3031, opened 2026-06-05
feat: add page size dropdown for topic data page

### LINKED ISSUE #3030: Page size dropdown for topic data page
On the topic list page, we have a dropdown to choose the page size 

Image

We should have the same on the topic data page, so the user choose how many records he/she wants to see by page

Image

Currently it is configurable only in the `application.yml`
https://github.com/tchiotludo/akhq/blob/626dfebd7fdc98119ab37e5b035ab5684e2b47cb/application.example.yml#L173-L174

Adding a new setting in the settings page to save the user preference + adding the dropdown on the topic data page could be a good idea

harshithasompura:
Closes #3030 

Added a dropdown for page size in the topic data

### DIFF
--- a/client/src/containers/Settings/Settings.jsx
+++ b/client/src/containers/Settings/Settings.jsx
@@ -15,6 +15,7 @@ class Settings extends Form {
       topicDefaultView: '',
       topicDataSort: '',
       topicDataDateTimeFormat: '',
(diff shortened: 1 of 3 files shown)


AlexisSouquiere:
Closed in #3143

PR #3032, opened 2026-06-06
feat(ui): show the current object id in the page title

### LINKED ISSUE #3026: Add more specific info into page title
I would like to ask you to add more specific info into page title for better orientation in browser tabs.

Now when you open particular topic to check messages, the page title is "Topics | akhq.io". When you have lot of opened topics in many browser tabs, then it is quiet hard to find one of them. It would be great when page title contains also topic name, e.g. "my-topic | Messages | akhq.io".

The same is for topic consumer groups view. When you open particular topic to check consumer groups, the page title is "Consumer Groups | akhq.io".  Would be nice to contain topic name, e.g. "my-topic | Consumer Groups | akhq.io".

gggeon96:
Includes the topic name in the page title on a topic page (e.g. "<topic> | Topics | akhq.io"), taken from the route, so topic tabs are easier to tell apart. The topic list page is unchanged.

Scoped to topics for now; connect/registry/etc. could follow as you mentioned.

Closes #3026

### DIFF
--- a/client/src/components/Base/Base.jsx
+++ b/client/src/components/Base/Base.jsx
@@ -31,31 +31,48 @@ class Base extends Root {
     };
   }
 
+  getTitleContext(pathname) {
+    const segments = pathname.split('/').filter(Boolean);
+    const section = segments[2] || '';
+    const decodeSegment = value => {
+      if (!value) return '';
+      try {
+        return decodeURIComponent(value);
+      } catch (e) {
+        return value;
+      }
+    };
+    switch (section) {
+      case 'topic':
+        return { label: 'Topics', id: decodeSegment(segments[3]) };
+      case 'node':
+        return { label: 'Nodes', id: decodeSegment(segments[3]) };
+      case 'tail':
+        return { label: 'Live Tail', id: '' };
+      case 'group':
+        return { label: 'Consumer Groups', id: decodeSegment(segments[3]) };
+      case 'acls':
+        return { label: 'Acls', id: decodeSegment(segments[3]) };
+      case 'schema': {
+        const schemaId = segments[3] === 'details' ? segments[4] : '';
+        return { label: 'Schema Registry', id: decodeSegment(schemaId) };
+      }
+      case 'connect': {
+        const connectId = segments[3];
+        const definitionId = segments[4] === 'definition' ? segments[5] : '';
+        return { label: 'Connect', id: decodeSegment(definitionId || connectId) };
+      }
+      case 'ksqldb':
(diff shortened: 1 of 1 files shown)


client/src/components/Base/Base.jsx
AlexisSouquiere, line 21: Because of topicId being empty here, your change doesn't show the topicId. We need a deeper rewrite of this component params evaluation

AlexisSouquiere (changes requested):
We need a deeper rewrite in this component. `getDerivedStateFromProps` doesn't work and doesn't return anything

To make it works for topic, connect, schema, etc. we need to retrieve the title context (including the object id like the topicId, subjectId, connectId, etc.). Something like

```ts
  getTitleContext(pathname) {
    const segments = pathname.split('/').filter(Boolean);
    const section = segments[2] || '';

    const decodeSegment = value => {
      if (!value) return '';
      try {
        return decodeURIComponent(value);
      } catch (e) {
        return value;
      }
    };

    switch (section) {
      case 'topic':
        return { label: 'Topics', id: decodeSegment(segments[3]) };
      case 'node':
        return { label: 'Nodes', id: decodeSegment(segments[3]) };
      case 'tail':
        return { label: 'Live Tail', id: '' };
      case 'group':
        return { label: 'Consumer Groups', id: decodeSegment(segments[3]) };
      case 'acls':
        return { label: 'Acls', id: decodeSegment(segments[3]) };
      case 'schema': {
        const schemaId = segments[3] === 'details' ? segments[4] : '';
        return { label: 'Schema Registry', id: decodeSegment(schemaId) };
      }
      case 'connect': {
        const connectId = segments[3];
        const definitionId = segments[4] === 'definition' ? segments[5] : '';
        return { label: 'Connect', id: decodeSegment(definitionId || connectId) };
      }
      case 'ksqldb':
        return { label: 'KsqlDB', id: decodeSegment(segments[3]) };
      default:
        return { label: '', id: '' };
    }
  }

  handleTitle() {
    const page = window.location.pathname;
    const { label, id } = this.getTitleContext(page);
    const title = label ? (id ? `${id} | ${label} |` : `${label} |`) : '';
    return title + ' akhq.io';
  }
```

client/src/components/Base/Base.jsx
gggeon96: Sorry, I hadn't realized the params aren't populated at this layout level, so reading topicId from state never worked. Reworked it to read the title context from the URL path like you suggested, now covering topic, node, group, acls, schema, connect and ksqldb. Pushed in `a064f6e5`.

AlexisSouquiere (approved):
LGTM !

PR #3041, opened 2026-06-12
fix(auth): preserve return route after session expiry

### LINKED ISSUE #3029: Previous page redirection after session expiration and re-login not working
After a session expiration, user is redirected to the login page. 
After login again, the user should go back to the previous page. This is not how it works today

JoshuaLutgendorff:
Fixes #3029.

This change preserves the current UI route when a user's session expires and they are redirected back to the login page.

Previously, when the JWT/session was lost, the app redirected to login but did not always save the route the user was on. After logging back in, the user could be sent to the default page instead of returning to their previous page.

This adds a session-expiry-specific return route save, while keeping the existing component unmount/logout return behavior unchanged.

Following review feedback, the saved returnTo route now includes basePath so the redirect remains consistent for context-path setups. The login redirect now uses the saved returnTo directly, or falls back to basePath + '/ui' when no return route exists.

I also wrapped AkhqRoutes with the existing withRouter helper so Root can access this.props.location?.pathname when saving the return route.

Tested locally by:

enabling basic auth
navigating to a protected page
deleting the JWT cookie
refreshing the page
confirming returnTo is saved without duplicate /ui
logging in again
confirming the app returns to the previous route and clears returnTo

Please let me know if this matches the expected behavior for the session-expiry flow, or if you would prefer the route-saving logic handled differently.

### DIFF
--- a/client/src/components/Root/Root.jsx
+++ b/client/src/components/Root/Root.jsx
@@ -1,4 +1,5 @@
 import { Component } from 'react';
+import { basePath } from '../../utils/endpoints';
 import axios from 'axios';
 import { get, put, post, remove } from '../../utils/api';
 
@@ -10,17 +11,17 @@ class Root extends Component {
     const pathname = this.props.location?.pathname;
 
     if (pathname && pathname !== '/ui/login') {
-      sessionStorage.setItem('returnTo', pathname + (window.location.search || ''));
+      sessionStorage.setItem('returnTo', basePath + pathname + (window.location.search || ''));
     }
 
     this.cancelAxiosRequests();
   }
 
   saveReturnToOnSessionExpiry() {
-    const pathname = window.location.pathname;
+    const pathname = this.props.location?.pathname;
 
     if (pathname && pathname.startsWith('/ui/') && pathname !== '/ui/login') {
-      sessionStorage.setItem('returnTo', pathname + (window.location.search || ''));
+      sessionStorage.setItem('returnTo', basePath + pathname + (window.location.search || ''));
     }
   }
 
--- a/client/src/components/Root/Root.jsx
+++ b/client/src/components/Root/Root.jsx
@@ -16,6 +16,14 @@ class Root extends Component {
     this.cancelAxiosRequests();
   }
 
+  saveReturnToOnSessionExpiry() {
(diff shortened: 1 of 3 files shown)


client/src/components/Root/Root.jsx
AlexisSouquiere:
```ts
import { basePath } from '../../utils/endpoints';

...

sessionStorage.setItem('returnTo', basePath + pathname + (window.location.search || ''));
```

client/src/components/Root/Root.jsx
```diff
@@ -16,6 +16,14 @@ class Root extends Component {
     this.cancelAxiosRequests();
   }
 
+  saveReturnToOnSessionExpiry() {
+    const pathname = window.location.pathname;
```
AlexisSouquiere: Use `this.props.location?.pathname` instead

client/src/components/Root/Root.jsx
```diff
@@ -16,6 +16,14 @@ class Root extends Component {
...
+
+    if (pathname && pathname.startsWith('/ui/') && pathname !== '/ui/login') {
+      sessionStorage.setItem('returnTo', pathname + (window.location.search || ''));
```
AlexisSouquiere:
```ts
sessionStorage.setItem('returnTo', basePath + pathname + (window.location.search || ''));
```

AlexisSouquiere (changes requested):
Hi @JoshuaLutgendorff ! Thank you for the PR. It works fine for the classic setup :)

The only thing is when a user setup a `context-path` (https://akhq.io/docs/configuration/others.html#server) there is an issue with the returnTo management (sometimes it contains the base path, sometimes it doesn't)

According to the tests I did on the 2 setups, we should be able to make it works by:
- Appending the basePath to the returnTo variable in the Root component (also in the componentWillUnmount)
- Changing a bit the behaviour after login. Can you update the Login.jsx on line 88 like this: `window.location.replace((returnTo || basePath + '/ui'))` ? We won't append anymore the basePath to `returnTo` because we already managed it before. We keep it only when there is `returnTo`

AlexisSouquiere (approved):

PR #3110, opened 2026-07-13
Add audit events for message produce, record delete and empty topic

Zanatti:
## What this does

Adds audit events for three topic-data actions that weren't audited yet: producing records, deleting a record, and emptying a topic.

## How

- New `RecordAuditEvent` (type `RECORD`), following the same pattern as `TopicAuditEvent` and the other audit events, and registered in `AuditEvent`'s `@JsonSubTypes`.
- Three new `ActionType` values: `RECORD_PRODUCE`, `RECORD_DELETE`, `RECORD_EMPTY`.
- `RecordRepository` emits the event after the Kafka operation succeeds, in `produce`, `delete`, `emptyTopic` and `emptyTopicByTimestamp`.
- Updated `docs/docs/configuration/audit.md`.

## A couple of choices worth flagging

Record keys and values are not logged, to keep payload data out of the audit topic. Only metadata goes in: cluster, topic, record count, and partition for deletes.

The produce event stores how many records were produced rather than a partition, since a multi-message produce can span partitions and the requested partition is usually empty. If you'd prefer per-record partition/offset there, I'm happy to add it.

## Tests

Extended `KafkaAuditModuleTest` to cover all four call sites against the embedded cluster: single produce, multi-message produce (checks the count), delete, `emptyTopic`, and `emptyTopicByTimestamp`. The audit tests pass.

### DIFF
--- a/docs/docs/configuration/audit.md
+++ b/docs/docs/configuration/audit.md
@@ -7,6 +7,10 @@ akhq can be configured to emit audit event to a kafka cluster for the following 
   - Topic configuration change
   - Topic partition increase
   - Topic deletion
+- Topic data level
+  - Produce records to a topic
+  - Delete a record
+  - Empty a topic
 - Consumer group level
   - Update offsets
   - Delete offsets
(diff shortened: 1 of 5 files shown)


samuelstuetz (approved):
lgtm

grigassukys:
This is very useful!

AlexisSouquiere (approved):
LGTM !

PR #3122, opened 2026-07-23
feat: allow searching schemas by numeric ID (#2450)

### LINKED ISSUE #2450: Schema registry search by schema id

Mubashir78:
## Summary

This PR adds the ability to search schemas by their numeric ID in the Schema Registry page, in addition to the existing subject name search.

## Problem

When users typed a numeric schema ID (e.g. `5`) in the Schema Registry search bar, the page showed "No schemas available" because the search only performed substring matching on subject names. Subject names (like `my-topic-value`) never contain the numeric schema ID.

## Solution

When the search term is a valid integer, the `all()` method in `SchemaRegistryRepository` now also queries the schema registry by schema ID using the existing `getSubjectsBySchemaId()` method. Results from both name-based and ID-based searches are merged and deduplicated.

### Changes

- **`SchemaRegistryRepository.java`**: Modified the `all()` method to:
  1. Collect name-matched subjects (existing behavior)
  2. If the search term is numeric, also look up subjects by schema ID
  3. Merge results using a `LinkedHashSet` to avoid duplicates
  4. Sort and return the combined list

## Testing

This is a backend-only change. The search is a filter on the existing schema list endpoint, so:
- Searching by name still works as before
- Searching by numeric ID now returns matching schemas
- Searching by non-numeric text continues to work as before
- Empty search returns all schemas as before

Fixes #2450

### DIFF
--- a/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
+++ b/client/src/containers/Topic/Topic/TopicData/TopicData.jsx
@@ -13,6 +13,7 @@ import {
   uriTopicsPartitions
 } from '../../../../utils/endpoints';
 import Pagination from '../../../../components/Pagination/Pagination';
+import PageSize from '../../../../components/PageSize/PageSize';
 import DatePicker from '../../../../components/DatePicker';
 import camelCase from 'lodash/camelCase';
 import constants, { SETTINGS_VALUES } from '../../../../utils/constants';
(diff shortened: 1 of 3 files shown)


Mubashir78:
Closing to separate commits. Page size dropdown (#3030) is already in PR #3121.

PR #3141, opened 2026-08-03
fix(avro): handle Utf8 keys in map deserialization

akmhmgc:
## Summary

Follow-up to #3120. When an Avro map field's schema does not carry `\"avro.java.string\": \"String\"`, Confluent's `KafkaAvroDeserializer` returns `Utf8` instances as map keys instead of `java.lang.String`. `AvroDeserializer.mapDeserializer` was declared as `Map<String, ?>` and cast the incoming Map to that type. The cast succeeded at compile time via erasure but blew up at runtime with:

```
java.lang.ClassCastException: class org.apache.avro.util.Utf8 cannot be cast to class java.lang.String
    at org.akhq.utils.AvroDeserializer.mapDeserializer(AvroDeserializer.java:142)
```

The failure was masked in `Record.convertToString` by a catch-all that falls back to `new String(payload)`, so the UI rendered raw Avro bytes as garbled text instead of surfacing the deserialization failure. Because `exceptions` on `Record` is `@JsonIgnore`, there was no visible signal in the UI or the HTTP response either.

This only manifests for schemas that omit the `avro.java.string=String` hint on map values (or on the enclosing map). That's a perfectly valid Avro construct that's common in schemas authored without Java-specific hints (e.g. by non-JVM producers).

## Fix

Widen `mapDeserializer` to accept `Map<?, ?>` and call `k.toString()` when inserting into the result map. `CharSequence` values were already handled correctly by `objectDeserializer`'s `STRING` branch.

## Tests

Adds `testMapWithUtf8Keys` which constructs a `GenericRecord` and inserts `Utf8` instances directly as map keys. Without the fix it fails with the `ClassCastException` above; with the fix it passes. All 35 tests in `AvroDeserializerTest` pass.

## Test plan

- [x] `./gradlew test --tests \"org.akhq.utils.AvroDeserializerTest\" -x startTestKafkaCluster` — 35/35 pass
- [x] Verified with a real production payload from a `map<union[string, null]>` schema that does not set `avro.java.string=String`: rendered as garbled text before, renders as JSON after

### DIFF
--- a/src/main/java/org/akhq/utils/AvroDeserializer.java
+++ b/src/main/java/org/akhq/utils/AvroDeserializer.java
@@ -55,7 +55,6 @@ public class AvroDeserializer {
             );
     }
 
-    @SuppressWarnings("unchecked")
     private static Object objectDeserializer(Object value, Schema schema) {
         LogicalType logicalType = schema.getLogicalType();
         Type primitiveType = schema.getType();
@@ -87,7 +86,7 @@ public class AvroDeserializer {
                 case UNION:
                     return AvroDeserializer.unionDeserializer(value, schema);
                 case MAP:
-                    return AvroDeserializer.mapDeserializer((Map<String, ?>) value, schema);
+                    return AvroDeserializer.mapDeserializer((Map<?, ?>) value, schema);
                 case RECORD:
                     return AvroDeserializer.recordDeserializer((GenericRecord) value);
                 case ENUM:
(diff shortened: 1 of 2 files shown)


akmhmgc:
@tchiotludo @AlexisSouquiere 
While verifying #3120 against real production traffic, I noticed the fix was incomplete — a related ClassCastException still leaves records garbled in the UI. This PR addresses the remaining case (details in the PR description).

I've already verified the fix against real production data in our environment: a record that previously rendered as garbled bytes now renders correctly as JSON in the akhq UI.

Would appreciate a review whenever you have time. Thanks!

AlexisSouquiere (approved):
Good catch !

PR #3155, opened 2026-08-04
Add support for regex in json topic filters

azuzunaga:
This PR adds support for regex in json filter topics.

Big enough Kafka clusters might contain too many topics to manage manually. One option is to generate the AKHQ configuration mechanically. Another option, explored in this PR, is to allow for regex matching and key merging.

This way, a catch-all filter can be used for common fields, augmented by specific topic filters. This simplifies security management and decouples topic creation from AKHQ configuration changes.

The bulk of this PR are doc and test updates, with the logic changes only taking up 39 lines.


src/main/java/org/akhq/configs/JsonMaskingFilter.java
```diff
@@ -9,6 +9,6 @@
 @Data
 public class JsonMaskingFilter {
     String description = "UNKNOWN";
-    String topic = "UNKNOWN";
+    String topic;
```
azuzunaga, line 12: To the best of my understanding, this was being used as a `null` placeholder. It worked before because this filter lowercased the lookup key while storing config topics verbatim, meaning even a topic `UNKNOWN` would be lowercased and then looked up, which would never match. Now, we compile and lookup, so a topic `UNKNOWN` would match.

If a filter is missing a topic we now skip the filter with a warning.

src/main/java/org/akhq/utils/JsonShowByDefaultMasker.java
```diff
@@ -29,7 +29,7 @@ public Record maskRecord(Record record) {
             if (!isJson(record)) {
                 return record;
             }
-            String topic = record.getTopic().getName().toLowerCase();
+            String topic = record.getTopic().getName();
```
azuzunaga, line 32: This is now done in `src/main/java/org/akhq/utils/JsonMasker.java`, with a `Pattern.CASE_INSENSITIVE` match.

src/main/java/org/akhq/utils/JsonMasker.java
```diff
@@ -1,34 +1,41 @@
...
+                Pattern.compile(filter.getTopic(), Pattern.CASE_INSENSITIVE),
+                filter.getKeys()
+            ))
+            .toList();
     }
 
     protected List<String> getKeysForTopic(String topic) {
-        return topicToKeysMap.getOrDefault(topic.toLowerCase(), Collections.emptyList());
+        return topicPatternToKeys.stream()
+            .filter(entry -> entry.getKey().matcher(topic).matches())
```
azuzunaga, line 40: Match the full entry, no implicit substring match. User retains the ability to use regex wildcards to opt into substring matching.

docs/docs/configuration/akhq.md
```diff
@@ -232,10 +244,18 @@ If you wish to mask data this way, you can:
...
 
-NOTES: Only one filter per topic is currently supported. If you are using `RecordNameStrategy` on a topic with multiple record types,
+The `topic` field in each filter is a case-insensitive Java regular expression that must match the entire topic name.
+For example, `topic: user-events-.*` will match topic names like `user-events-login` and `USER-EVENTS-AUDIT`, but `topic: users` will not match `users-archive`.
+When multiple filters match the same topic, the keys from all matching filters are combined, and the order of filters does not matter.
+
+In earlier AKHQ versions, `topic` was matched as an exact string.
+When upgrading, escape any regex special characters in literal topic names (for example `orders\.v1` instead of `orders.v1`), otherwise the unescaped `.` matches any character and the filter may also apply to unintended topics.
+A filter without a `topic` is ignored (with a warning in the logs).
+A `topic` that is not a valid regular expression fails with a `PatternSyntaxException` when the masker is created.
```
azuzunaga: Configurations written before regex support cannot trigger this failure: Kafka topic names can only contain letters, digits, `.`, `_` and `-`, and all of those are syntactically valid regular expressions, so only new patterns can be malformed. Meaning, upgrading versions is backwards-compatible.

azuzunaga:
@tchiotludo, @AlexisSouquiere, let me know what you think of this approach. I didn't find a contributors guide, so I've done my best to explain my rationale and called out specific parts of the code that I thought merited an explanation.

@jamfor999, you might be interested in taking a look as well, since this expands on the work you did previously.

azuzunaga:
Hi folks, bumping this PR, thanks!

jamfor999:
Personally, I think for mask by default at least there should be a config option to turn this behaviour on. This mode would usually be used as a safety feature for a production cluster with sensitive information, and adding wildcard/regex support undermines that a little.

For example to give some stakeholder perspective - in my org we mechanically build the AKHQ config for the fields to unmask. Each team has its own subfolder with CODEOWNERS, where they fill out their own info, and we have checks to make sure each team can only unmask fields on topics they own. The ability to use wildcards is something we would _never_ want at an org level because it could match too many things.

Our org could, of course, add another custom PR check to make sure nobody uses a wildcard, but if AKHQ is going to support them, it feels like it should also have a config option to disable them completely at AKHQ level rather than orgs working around it. Otherwise, IMO, this should be considered a breaking change because now the understanding of what unmasks what has changed.

Could you add something like enable-regex-topic-filters, with tests for when it’s on and off? When it’s off, topic filters should keep using exact matches, which imo should be the default. 

Obviously at the end of the day I don’t own AKHQ, so it’s ultimately up to the owners what should go in. I just thought I’d give my two cents as a stakeholder and explain the original thinking behind all this.

azuzunaga:
@jamfor999, good point and always interesting to hear about other use cases. Pushed some changes, let me know if that's what you had in mind.

PR #3227, opened 2026-09-07
fix(ui): support Unicode ACL principals

### LINKED ISSUE #1920: ACLs screen fails to load if subjects contain unicode characters
If the attached Kafka cluster has ACL configuration containing certificates with Unicode characters in the subject field, AKHQ completely fails to load the ACLs screen. This is because `btoa` used does not support non-latin characters.

see: https://developer.mozilla.org/en-US/docs/Web/API/Window/btoa#unicode_strings

AKHQ version: 0.25.1
```
Uncaught (in promise) DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.
    at https://<domain>/ui/assets/index-BdxSfnGT.js:476:54779
    at Array.map (<anonymous>)
    at x5.handleData (https://<domain>/ui/assets/index-BdxSfnGT.js:476:54752)
    at x5.getAcls (https://<domain>/ui/assets/index-BdxSfnGT.js:476:54711)
```

aoto-tech:
## Summary

- encode ACL principals as UTF-8 before Base64 conversion
- preserve the existing Base64 representation while escaping it when used in URL path segments
- add regression coverage for Unicode principal values and path-unsafe Base64 output

## Problem

Calling `btoa()` directly with a Unicode principal throws a `DOMException`, preventing the ACL list from loading. Standard Base64 can also contain `/` and `+`, so the UI now escapes those values only at URL path boundaries without changing the Base64 alphabet expected by existing API code.

## Testing

- `npm test -- --run src/utils/base64.test.js src/containers/Acl/AclList/Acls.test.jsx`
- `npm run build`
- `JAVA_HOME=/home/ubuntu/.local/jdks/jdk-25.0.4.1+1 PATH=/home/ubuntu/.local/jdks/jdk-25.0.4.1+1/bin:$PATH ./gradlew test --tests org.akhq.models.AccessControlTest -x distTar -x distZip -x shadowDistTar -x shadowDistZip -x buildLayers -x shadowJar -x runnerJar`
- `git diff --check`

Notes: full frontend lint/test/prettier checks were not clean in this local environment due existing unrelated failures/configuration issues.

Fixes #1920

### DIFF
--- a/client/src/containers/Acl/AclDetail/AclDetails.jsx
+++ b/client/src/containers/Acl/AclDetail/AclDetails.jsx
@@ -8,27 +8,36 @@ import { getSelectedTab } from '../../../utils/functions';
 import { Link } from 'react-router-dom';
 import Root from '../../../components/Root';
 import { withRouter } from '../../../utils/withRouter';
+import {
+  decodeBase64PathSegment,
+  decodeUtf8FromBase64,
+  encodeBase64PathSegment
+} from '../../../utils/base64';
 
 class AclDetails extends Root {
   state = {
     clusterId: this.props.params.clusterId,
-    principalEncoded: this.props.params.principalEncoded,
+    principalEncoded: decodeBase64PathSegment(this.props.params.principalEncoded),
     selectedTab: 'topics'
   };
 
   tabs = ['topics', 'groups', 'clusters', 'transactionalids'];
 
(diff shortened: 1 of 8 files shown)


aoto-tech:
wait Reviewers status!

PR #3260, opened 2026-09-21
fix(ui): normalize state handling to uppercase in ConsumerGroupList and TopicGroups

### LINKED ISSUE #3248: [UI/UX] - Consumer Groups status stable is a misleading colour
# Issue description

Since the latest version of AKHQ the status label in the Consumer Group page has changed colour from a green to a yellow. 

Image

The yellow colour is misleading as the basis of UI/UX tells us that yellow/orange means warning/requires attention.

# Expected behaviour

I propose to change the color of the stable consumer group from yellow to the same green that can be found in the connector page `--bs-success-rgb`

Image

# Fix 

The issue lies in the Frontend in `handleState` in [`ConsumerGroupList.jsx`](https://github.com/tchiotludo/akhq/blob/602d69df7f8687d1036e1520df6f285b20139eda/client/src/containers/ConsumerGroup/ConsumerGroupList/ConsumerGroupList.jsx#L127C2-L127C14) and in `handleState` in [`TopicGroups.jsx`](https://github.com/tchiotludo/akhq/blob/602d69df7f8687d1036e1520df6f285b20139eda/client/src/containers/Topic/Topic/TopicGroups/TopicGroups.jsx#L91) where the map between the Consumer Group state & its colour is case sensitive.

# Why 

Starting with Kafka 4 the [`ConsumerGroupState`](https://kafka.apache.org/41/javadoc/org/apache/kafka/common/ConsumerGroupState.html) enum that describe the state in which a consumer group is (`STABLE`,`EMPTY, `etc...) has been deprecated in favour of the [`GroupState`](https://kafka.apache.org/41/javadoc/org/apache/kafka/common/GroupState.html) enum that be found [here](https://github.com/apache/kafka/blob/849534dc7cb0134ebc124c3bb4fad6e03e15740c/clients/src/main/java/org/apache/kafka/common/GroupState.java#L51) now states are reported (`Stable`,`Empty`, etc...). That cause both `handleState` in the frontend to fall back to the default `--bs-warning-rgb`

nil-malh:
# Fix

Kafka’s consumer-group API changed the format used to report states: instead of the previous enum-style value such as  PREPARING_REBALANCE/STABLE , newer Kafka versions can return a name such as  PreparingRebalance/Stable. The backend forwards Kafka’s reported state directly, so the frontend’s exact string comparisons stopped matching.

The commit normalizes received states with  state.toUpperCase()  and updates the comparison to  PREPARINGREBALANCE/STABLE , allowing both casing formats to be handled while preserving the original label shown to users. It applies this to:

- Consumer group list: correctly styles the  PreparingRebalance  state.
- Topic group list: continues recognizing  Stable  regardless of casing.

In short, this is a compatibility fix for Kafka changing how it serializes/reports consumer-group status, not a change to the consumer-group behavior itself.

Resolves #3248 

# Proof of  tests

Created a simple consumer group that consumes a topic to report a 

screenshot_20260921_212006


AlexisSouquiere (approved):

AlexisSouquiere:
Thank you @nil-malh !

